PR #41, opened 2019-01-21
Docs/nunjucks macros

davidhunter08:
Updated the How to guides...
- The Adding Nunjucks page has now become 'How to use components'
- The Writing CSS page has now become 'How to style page elements'
Content for both and URLs for both these pages have been updated.


chrimesdev (approved):
Content changes 👍

PR #51, opened 2019-01-30
Moved templates and examples into the app folder

davidhunter08:
The examples and templates have been moved to the app folder from the docs folder.
Users were getting confused about where to put the templates.


bencullimore (approved):
👌

PR #52, opened 2019-02-04
Added template.env file to lib folder

davidhunter08:
A template.env file was missing from the lib folder, breaking Heroku builds.


bencullimore (approved):
👍

Issue #62, opened 2019-02-11
Broken links on mini hub page template


davidhunter08:
See http://nhsuk-prototype-kit.azurewebsites.net/templates/mini-hub

chrimesdev:
Might be worth doing a CTRL+F for any `/docs/` page links as most of them have seemed to have moved to `/app/`

PR #63, opened 2019-02-12
[refs #62] Fix broken links

mcheung-nhs:
After files moved from /docs/, these pages still had references to that
location.


chrimesdev (approved):
Looks good

PR #96, opened 2019-03-13
Moved examples and templates to from app to docs

davidhunter08:
- moved examples and templates from app to docs
- updated documentation_routes.js
- deleted layout.html from app folder
- updated templates, passing data and branching guidance


app/views/content-page.html
```diff
@@ -1,7 +1,7 @@
 {% extends 'layouts/layout-nhsuk.html' %}
```
chrimesdev: There seems to be 2 `content-page` with different content

One at `/templates/content-page` (docs/views/templates/content-page.html) which is linked to from the page templates and one at `/templates/content-page` (app/views/content-page) which is linked to from the non-docs website homepage - this page seems to have better content. They used to have the same content but have diverged, do we need both?

app/views/includes/scripts.html
```diff
@@ -1,5 +1,4 @@
 <script src="/js/main.js"></script>
 <script src="/js/jquery-3.3.1.min.js"></script>
-<script src="/js/auto-store-data.js"></script>
```
chrimesdev, line 3: Is this the old method of passing data? Has it been replaced by another method?

docs/views/templates.html
```diff
@@ -18,7 +18,7 @@
 
       <h1>Page templates</h1>
 
-      <p>You can find these templates in the <strong>app/views/templates</strong> folder.</p>
+      <p>You can find these templates in the <strong>doc/views/templates</strong> folder.</p>
```
chrimesdev: Should be `docs` not `doc`

chrimesdev (changes requested):
Noticed a few changes needed.

Also need to bump to version in `package.json` and `package-lock.json`, with a CHANGELOG entry with an explanation of the changes.

app/views/index.html
```diff
@@ -28,19 +28,36 @@ <h1>
...
+      </h2>
 
-      <p>Use this page as the index for your project.</p>
+      <p>
+        Use this page as the index for your project.
+      </p>
       
-      <p>You'll find this page in the kit at <strong>/app/views/index.html</strong>.</p>
+      <p>
+        The code for this page in the kit, you can find it at <strong>/app/views/index.html</strong>.
```
chrimesdev: Does this sentence read properly?

chrimesdev: For consistency... when telling users to look in a folder, we've used '/' before the folder here, however on other examples such as on the `docs/views/how-tos/passing-data-page.html` page. There is no '/' before the folder 'docs'.

docs/views/how-tos/branching.html
```diff
@@ -21,7 +21,7 @@ <h1>How to do branching</h1>
 
       <p>The code for the example can be found in:</p>
 
-      <pre class="app-pre"><code class="app-code">/documentation_routes.js<br>/views/examples/branching</code></pre>
+      <pre class="app-pre"><code class="app-code">/documentation_routes.js<br>docs/views/examples/branching</code></pre>
```
chrimesdev, line 24: See comment here: https://github.com/nhsuk/nhsuk-prototype-kit/pull/96/files#r266793931

app/views/content-page.html
davidhunter08: We need both, I have renamed the one in app to page-example.html to save confusion.

docs/views/templates.html
davidhunter08: Done

app/views/index.html
davidhunter08: I have updated the content.

davidhunter08: I have added a '/' for all.

chrimesdev (approved):

PR #97, opened 2019-03-18
Added 'Updating the kit' guidance

davidhunter08:


docs/views/how-tos/updating-the-kit.html
```diff
@@ -0,0 +1,69 @@
...
+      <div class="nhsuk-inset-text">
+        <p>
+          If you have made any changes outside the <strong>app</strong> folder, this process will destroy those changes.
+          We will try and improve the update process to avoid this, but in the meantime you will need to make a note of your changes outside <strong>app</strong>, and add them back after updating.
+        </p>
+      </div>
+
+      <ol>
+        <li>
+          <a href="/docs/install/download-zip">Download the latest prototype kit</a>
```
chrimesdev: Does this need to say "Download the latest version of the prototype kit"?

Will users get confused thinking there's a new kit altogether?

docs/views/how-tos/updating-the-kit.html
```diff
@@ -0,0 +1,69 @@
...
+        <li>
+          Copy the <strong>config.js</strong> file from the app folder in the latest kit, to the <strong>app</strong> folder in your prototype.
+          <img class="app-img-guide nhsuk-u-margin-bottom-0 nhsuk-u-margin-top-4" alt="Screenshot of the config files in the latest prototype kit foldetr" src="/images/updating-the-kit/paste.png" />
+          <img class="app-img-guide nhsuk-u-margin-bottom-0" alt="Screenshot of the config files in the latest prototype kit foldetr" src="/images/updating-the-kit/paste-config.png" />
+          <div class="nhsuk-inset-text nhsuk-u-margin-top-0">
+            <p>If you've made any changes to the <strong>config.js</strong> file in your prototype then you'll need to re-enter them in the new version of the file e.g. the service name.</p>
+          </div>
+        </li>
+        <li>
+          Check <strong>\app\assets\sass\components</strong> in the latest kit for any new components. Copy the files over to your prototype.
```
chrimesdev: We use backslashes here `\` for folder paths but on the rest of the documentation we use forward slashes '/'.

Also does this need to be syntax highlighted with `<code class="app-code">` like we do elsewhere?

docs/views/how-tos/updating-the-kit.html
```diff
@@ -0,0 +1,69 @@
...
+          <img class="app-img-guide nhsuk-u-margin-bottom-0 nhsuk-u-margin-top-4" alt="Screenshot of the latest prototype kit folder" src="/images/updating-the-kit/components.png" />
+        </li>
+        <li>
+          Check <strong>\app\assets\sass\main.scss</strong> in the latest kit to see if any changes have been made in the top section, above where it says <strong>// Add your custom CSS/Sass styles below...</strong> .
```
chrimesdev: See backslashes vs forward slashes above.

chrimesdev (changes requested):
Should the "I already have the kit installed and want to download a new copy" option on the Install page (/docs/install) link through to this new guidance page?

docs/views/how-tos/updating-the-kit.html
davidhunter08: Updated to say "Download the latest version of the prototype kit"

docs/views/how-tos/updating-the-kit.html
davidhunter08: Updated to "/"

docs/views/how-tos/updating-the-kit.html
davidhunter08: Updated to "/"

davidhunter08: We're only using the syntax highlight for code and not for things like file name or file location.

PR #102, opened 2019-03-22
update main element ID from main-content to maincontent

chrimesdev:
Inline with the guidance of the NHS digital service manual, the
ID of the main element should be a unique ID of maincontent (rather
than main-content).
https://beta.nhs.uk/service-manual/styles-components-patterns/skip-link


bencullimore (approved):
I approve on the behalf of David (who's locked out of github)

PR #134, opened 2019-10-07
Create LICENSE

chrisns:
A license would be great to allow other people to reuse this code
used MIT since @nhsuk use it elsewhere, if it doesn't fit please suggest an alternative and I'll amend this PR


chrimesdev (approved):
This looks fine, thanks @chrisns 👍

PR #147, opened 2020-03-26
WIP - Add static site generation task

tomdoughty:
## Description
Added a task to generate a static site rather than a site requiring a Node server

### Todo
Build JS from nhsuk-frontend node module
Add documentation on how to use

## Checklist

- [ ] CHANGELOG entry


frankieroberto:
I'll close this for now, as is 4 years old, but it's interesting idea.

Rendering a static site would mean easier/cheaper hosting, but would also mean you can't do any complex routing on using form data in a 'check your answers' type page.

PR #148, opened 2020-04-17
:heavy_minus_sign: Removed the showLogo attribute from the footer block

mcheung-nhs:
## Description
There's no showLogo attribute in the footer macro. It was probably copied from the header macro? So I've removed it.

## Checklist

- [ ] CHANGELOG entry


davidhunter08 (approved):
LGTM. I think it was leftover from when we changed the design of the footer (making it grey etc).

chrimesdev:
Looks fine to me, I think it just needs a version bump and change log

PR #150, opened 2020-05-23
WIP: Form validation with express-validator

tomdoughty:
## Description
The guidance is written in `docs/views/how-tos/validation`. 

To use this you only need to add a rule to the `validation-rules` file using the same key you would use for existing data functionality.

- Added express-validator package to application
- Added middleware to run validator rules
- Added middleware to add errors to app state
  - `errorsList` is compatible with error summary component
  - `errorsObj` allows you to access specific errors for input error messages
- Middleware redirects to referer URL rather than post URL if errors exist
- Written basic guidance

## TODO
Confirm there is a need for this in the kit with users and contributors.

## Checklist

- [ ] CHANGELOG entry


chrimesdev:
I've recently implemented this into a prototype and it works well. Here are some notes (some of these steps are already documented and I am using vanilla HTML rather than Nunjucks)

- Added `{{ "Error: " if hasErrors }} ` to the pageTitle block in the layout file - I wonder if we could do this in the templates in the prototype and remove the need for users to have to do this step?
- Added the Error summary block below the `<form>` using Nunjucks example provided - need to look at how to do this in HTML and loop through the errors in HTML
- Added `{% if hasErrors %} nhsuk-form-group--error{% endif %}` to `nhsuk-form-group`
- Added `{% if hasErrors %} aria-describedby="error-example-id"{% endif %}` to the `fieldset` 
- Added the individual Error message to the form - need to look at how to pass the error messages to the text rather than hard coding the message:

```
{% if hasErrors %}
  <span class="nhsuk-error-message" id="error-example-id">
    <span class="nhsuk-u-visually-hidden">Error:</span> Select if you need to leave your home to go to work
  </span>
{% endif %}
```

I had to make sure the ID of the first form fields matched the name field, to make the sure the JavaScript focus worked.

frankieroberto:
Closing this for now, as it’s 4 years old and conflicts with main.

Would be good to explore different approaches to implementing validation at some point - although often this isn’t needed for prototypes.

PR #203, opened 2021-08-18
Swap outdated panel for card component

### LINKED ISSUE #184: The clear data page doesn't render in the passing data how to guide
# Bug Report

## What is the issue?
When going through the passing data example, if you click on the "Clear data" link in the footer, you get a page error.  

1. Go to: https://nhsuk-prototype-kit.azurewebsites.net/examples/passing-data/passing-data-question-name
2. Click on the "Clear data" link in the footer
3. Observe the error which includes: "Unable to call 'panel', which is undefined or falsey"

* Device: all
* Operating System: all
* Browser: all

The promo component is no longer in use in the NHSUK frontend. Need to replace it with the card component.

magicaljellybeans:
## Description

Use of `panel` on `clear-data.html` page was causing #184.
Swapped for `card` component and now the page is working.
 
## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # NHS.UK prototype kit Changelog
 
+## 4.6.1 - 18 August 2021
+
+:wrench: **Fixes**
+
+- Swap outdated `panel` for `card` on `clear-data.html` page.
+
 ## 4.6.0 - 16 August 2021
 
 :new: **New features**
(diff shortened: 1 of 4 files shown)


chrimesdev:
Good spot, looks good. Just needs version number bumping in the `package.json` files and the `"headingLevel"` should be 1 for the `<h1>` of the page

chrimesdev (approved):

PR #211, opened 2021-08-25
remove release action for uploading versioned assets

chrimesdev:
## Description

We can remove the action to upload the `zip` file with the version number in the name, as there is no compiled `zip` folder to upload from the project and we are just going to use the default `zip` file that gets uploaded during a release. eg. https://github.com/nhsuk/nhsuk-prototype-kit/archive/refs/tags/v4.6.3.zip


DomBaker (approved):
reviewed and approved

PR #285, opened 2024-05-15
Remove .vs folder

frankieroberto:
Looks like this Visual Studio local folder was accidentally committed.


anandamaryon1 (approved):
Thanks for tidying, let's get it in the bin 🗑️

PR #224, opened 2022-01-25
Remove unnecessary console.log and fix js linting warnings

mcheung-nhs:
## Description
Remove unnecessary console.log and fix js linting warnings

## Checklist

- [ ] Tested against our [testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
- [x] CHANGELOG entry

PR #289, opened 2024-05-21
Remove unnecessary console.log

frankieroberto:
Currently all the session data keys are logged to the console on every page load. This removes that, as it’s unnecessary and confusing.

Previously suggested in #224


edwardhorsford (approved):
Thanks! this has been really bugging me.

PR #290, opened 2024-05-21
Add dependabot

frankieroberto:
This should help us keep the dependancies of the NHS Prototype Kit more up to date.


frankieroberto:
@roshaanbajwa @edwardhorsford does this seem sensible? I noticed that some of the dependancies are out-of-date.

The config can be tweaked if needed to ignore any major updates with breaking changes that we’re not yet ready for.

roshaanbajwa (approved):

PR #291, opened 2024-05-21
Ignore eslint warnings

frankieroberto:
This ignores some existing eslint (javascript style) warnings currently present in `main`:

* Unexpected unnamed function (x10)
* Unexpected console statement  (x4)

This gets us to zero warnings, so that we could enable it as a required check before merging.

Thanks to @mcheung-nhs for originally suggesting in this #224.

### DIFF
--- a/app.js
+++ b/app.js
@@ -102,7 +102,7 @@ if (useAutoStoreData === 'true') {
 function checkFiles() {
   const nodeModulesExists = fs.existsSync(path.join(__dirname, '/node_modules'));
   if (!nodeModulesExists) {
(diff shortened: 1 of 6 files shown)


edwardhorsford (approved):

joelanman:
just wondering if its easier/more robust to use a config file?

https://eslint.org/docs/latest/use/configure/

frankieroberto:
@joelanman probably we need to have a whole conversation about listing and how to set it up... 

I’d be in favour of us switching to [StandardJS](https://standardjs.com) (as used by GDS and others). Currently we’re using the [AirBnB config](https://www.npmjs.com/package/eslint-config-airbnb-base).

PR #300, opened 2024-05-23
Stop watching html files

### LINKED ISSUE #287: Session data is lost whenever a template file is saved
# Bug Report

Currently, whenever you edit a template, the kit restarts itself, which results in any session data being lost.

## What is the issue?

Session data should persist even when the template files are edited.

Set up a page with a form, and then display the information from the form on the next page using `{{ data | dump }}`.

Then edit a view template, and the page will reload with the data missing.

* Device: Mac
* Browser: Safari

frankieroberto:
Currently files with the extension `.html` are being watched, which means that whenever they are edited, the nodemon server is restarted.

When combined with the in-memory session storage (which is currently the default), this means that when editing any html file, all data previously entered by the user is wiped. This makes local development of prototypes where data is stored quite tricky.

Since nunjucks is now configured to not cache the files (see #286), we no longer need to watch the HTML files and reload the server, as any changes to the content will still be shown.

Fixes #287

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Disable nunjucks caching so changes to templates show immediately
+- Stop watching .html files and restarting the server whenever they are changed
 
 ## 4.10.0 - 22 February 2024
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -62,7 +62,7 @@ function startNodemon(done) {
   const server = nodemon({
     script: 'app.js',
     stdout: false,
-    ext: 'scss js html',
+    ext: 'scss js',
     quiet: true,
   });
   let starting = false;


frankieroberto:
Be great if someone else could test this and verify it fixes the issue for them.

To recreate the bug, I added a view file in `app/test.html` with the following content:

```nunjucks
{{ data | dump }}

<form action="/" method="post">
  {{ input({
    "label": {
      "text": "Full name"
    },
    "id": "name",
    "name": "name"
  }) }}

  {{ button({
    "text": "Save and continue"
  }) }}
</form>
```

Steps to reproduce the issue:
1. Add a name in the field, press continue
2. See that name is shown on the page
3. Edit the `test.html` file
4. Reload the page.
5. See that name is no longer shown.

With the change in this PR, at step 5 the name should still be visible.

edwardhorsford (approved):
Awesome! Tested on my local prototype and data is preserved after editing a template.

joelanman:
you dont need to watch `scss` with nodemon either - there's another gulp task for that

we only watch js, json and env

https://github.com/alphagov/govuk-prototype-kit/blob/main/lib/dev-server.js#L111-L113

frankieroberto:
@joelanman:
> you dont need to watch `scss` with nodemon either - there's another gulp task for that
> 
> we only watch js, json and env
> 
> https://github.com/alphagov/govuk-prototype-kit/blob/main/lib/dev-server.js#L111-L113

Well spotted. Spun out into a separate PR: #302

PR #301, opened 2024-05-23
Update node version in setup-node step

frankieroberto:
We should be using node 20 as that's what the kit requires.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,9 +15,9 @@ jobs:
         with:
           persist-credentials: false
 
-      - uses: actions/setup-node@v1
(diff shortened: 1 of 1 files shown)


edwardhorsford (approved):
Related, we should add a `.nvmrc` file to the root with the version.

frankieroberto:
> Related, we should add a `.nvmrc` file to the root with the version.

Possibly! Although that only works with Node Version Manager (NVM) which is separate from Node and “other version managers are available”™️.

That said, it’s probably the most popular one and would be helpful - though need to make sure it’s kept in sync with the node version in `package.json`.

PR #310, opened 2024-06-06
Replace keypather with lodash

frankieroberto:
keypather is quite old and has some security issues with one of its dependencies. This replaces it with a function from lodash, which is already installed anyway.

Re-applies same update from the GOV.UK Prototype Kit here: https://github.com/alphagov/govuk-prototype-kit/pull/1155

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,5 +1,5 @@
 // NPM dependencies
-const getKeypath = require('keypather/get');
+const { get: getKeypath } = require('lodash');
 const path = require('path');
 
 // Require core and custom filters, merges to one object
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,7 @@
         "gulp-nodemon": "^2.5.0",
         "gulp-rename": "^2.0.0",
         "gulp-sass": "^5.1.0",
-        "keypather": "^3.1.0",
+        "lodash": "^4.17.21",
         "nhsuk-frontend": "^8.1.1",
         "nunjucks": "^3.2.4",
         "path": "^0.12.7",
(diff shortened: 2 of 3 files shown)


lib/utils.js
```diff
@@ -1,5 +1,5 @@
 // NPM dependencies
-const getKeypath = require('keypather/get');
+const { get: getKeypath } = require('lodash');
```
joelanman, line 2: this is assigning to `get` but the code later still refers to `getKeyPath`

frankieroberto: Must admit I was unfamiliar with this syntax, but in the previous change on the govuk kit (see https://github.com/alphagov/govuk-prototype-kit/pull/1155/files#diff-5dfe38baf287dcf756a17c2dd63483781b53bf4b669e10efdd01e74bcd8e780aR38) the `getKeypath` line was also left unchanged, so I assumed it was importing the `get` function from lodash but assigning it to `getKeyPath`?

frankieroberto: ah yes, looks like it's using this syntax to destructure the `get` function from lodash and assign it to the `getKeypath` local variable? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#assigning_to_new_variable_names

joelanman: gosh thats confusing syntax!

frankieroberto: heh, yes! Probably this could be refactored to be less confusing, but it felt slightly safer to stick to the exact changes made on the GOVUK prototype kit for now.

edwardhorsford (approved):
Tested with single property, dot notation, and square bracket notation.

I'm not entirely sure about keeping the `getKeypath` name, but's probably not a big deal.

PR #317, opened 2024-06-18
Update docs to use v20 of node

edwardhorsford:
## Description
Updates the install instructions to use v20 of Node rather than v8. I have not updated the screenshots.


frankieroberto:
@edwardhorsford would you mind updating the screenshots (at least the Mac ones)? We'll have to find a Windows user for the other ones...

Think we should probably update the screenshots at the same time as the body text to avoid confusion?

roshaanbajwa (dismissed):
Looks good to me

the-nathan-smith (approved):

frankieroberto (approved):

PR #324, opened 2024-07-08
Replace basic authentication with a password page and a cookie.

### LINKED ISSUE #323: Replace basic authentication with a custom password page
We (Record a vaccination team) have done some user research with NHS staff on a prototype, and experienced the issue where corporate IT had turned off basic authentication in Windows Edge, and so were unable to access the prototype.

The GOV.UK Prototype Kit team experienced similar, and replaced basic authentication with a custom password screen in this pull request: https://github.com/alphagov/govuk-prototype-kit/issues/1019

We should do similar?

frankieroberto:
## Description

This replaces the "basic authentication" (a browser popup asking for a username and password) with a redirect to a page asking users to enter a password.

As well as being more user-friendly, this also helps avoid an issue where some corporate networks have turned of basic authentication in the Edge browser.

The code here heavily borrows from the work done by the GOV.UK Prototype Kit team in https://github.com/alphagov/govuk-prototype-kit/pull/1182

Fixes #323

## Screenshot

Screenshot of a page saying 'This is a prototype used for research. It is not a real service. You should only continue if you have been invited to test this prototype.' followed by a password box and a Continue button.

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/docs/views/how-tos/publish-your-prototype-online.html
+++ b/docs/views/how-tos/publish-your-prototype-online.html
@@ -35,7 +35,7 @@
       <h2>Setting a password</h2>
       <p>When running the prototype kit online, you must set a password. This is to stop anyone accidentally finding your prototype and mistaking it for a real service.</p>
       <p>To set a password, check your hosting services documentation on how to set "environment variables". (It may have a slightly different name like "config vars" or "variables".)</p>
-      <p>The password should be set using the variable <code class="app-code">PROTOTYPE_USERNAME</code>.</p>
+      <p>The password should be set using the variable <code class="app-code">PROTOTYPE_PASSWORD</code>.</p>
 
       <p>If you are using an older version of the prototype kit (before v4.12.9) you will also need to set a username using the variable <code class="app-code">PROTOTYPE_USERNAME</code>.</p>
 
--- a/docs/views/how-tos/publish-your-prototype-online.html
+++ b/docs/views/how-tos/publish-your-prototype-online.html
@@ -33,13 +33,13 @@
       <p>Some hosting services may automatically deploy your prototype every time you merge new changes in the connected GitHub repository. You may be able to enable automatic deploys. If it hasn't automatically update your prototype, you will need to do this manually.</p>
       
       <h2>Setting a password</h2>
-      <p>When running the prototype kit online, you must set a username and password. This is to stop anyone accidentally finding your prototype and mistaking it for a real service.</p>
+      <p>When running the prototype kit online, you must set a password. This is to stop anyone accidentally finding your prototype and mistaking it for a real service.</p>
       <p>To set a password, check your hosting services documentation on how to set "environment variables". (It may have a slightly different name like "config vars" or "variables".)</p>
-      <p>Set environment variables. For example:</p>
-      <p><code class="app-code">PROTOTYPE_USERNAME</code><code class="app-code">nameofservice</code></p> 
-      <p><code class="app-code">PROTOTYPE_PASSWORD</code><code class="app-code">yourpassword123</code></p>
+      <p>The password should be set using the variable <code class="app-code">PROTOTYPE_USERNAME</code>.</p>
 
-      <p>In Railway, you also need to create:</p> 
+      <p>If you are using an older version of the prototype kit (before v4.12.9) you will also need to set a username using the variable <code class="app-code">PROTOTYPE_USERNAME</code>.</p>
+
+      <p>In Railway, you also need to create:</p>
       <ul>
         <li>a variable called <code class="app-code">NODE-ENV</code> and set the value to <code class="app-code">production</code></li>
         <li>a variable called <code class="app-code">USE_AUTH</code> with the value of <code class="app-code">true</code></li>
--- a/docs/views/how-tos/publish-your-prototype-online.html
+++ b/docs/views/how-tos/publish-your-prototype-online.html
@@ -37,7 +37,7 @@
       <p>To set a password, check your hosting services documentation on how to set "environment variables". (It may have a slightly different name like "config vars" or "variables".)</p>
       <p>The password should be set using the variable <code class="app-code">PROTOTYPE_PASSWORD</code>.</p>
 
-      <p>If you are using an older version of the prototype kit (before v4.12.9) you will also need to set a username using the variable <code class="app-code">PROTOTYPE_USERNAME</code>.</p>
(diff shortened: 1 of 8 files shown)


frankieroberto:
@anandamaryon1 @edwardhorsford @paulrobertlloyd fancy testing this out, including deploying it some hosting environments like Heroku?

frankieroberto:
Possibly the password environment variable should be `PROTOTYPE_PASSWORD` not `PASSWORD`, for backwards-compatibility with the current kit? 🤔

anandamaryon1:
Looks like this how-to guidance will need to be updated to reflect this work: https://nhsuk-prototype-kit.azurewebsites.net/docs/how-tos/publish-your-prototype-online

frankieroberto:
@anandamaryon1 yep! And I think we would _also_ have to maintain guidance on how to set the username and password on previous versions (as not everyone will upgrade quickly?)

In which case, I think we could do something like this:

> When running the prototype kit online, you must set password to stop anyone accidentally finding your prototype and mistaking it for a real service.
>
> To do this, set an environment variable (called ‘Config vars’ on Heroku):
>
> `PROTOTYPE_PASSWORD=yourpassword123`
>
> If you are using an older version of the NHS Prototype Kit (before x.x) then you will also need to set a username. This can be the same of your service. For example:
> 
> `PROTOTYPE_USERNAME=nameofservice`

sarawilcox:
I'll see if I can find some designers who might be able to test this.

Tosin-Balogun:
@frankieroberto thank you for this. Just to save me a bit of time, are you able to share a heroku or github page preview so I can see how it works?

frankieroberto:
@Tosin-Balogun @sarawilcox @anandamaryon1 hiya. Sorry, I was on holiday so didn't see this update straight away.

I've deployed a quick demo of the password page here: https://nhsuk-prototype-password-page-b7626ee05c0c.herokuapp.com/ (password `testing`).

You can also test it locally, by adding these to a `.env` file:

```
PASSWORD=testing
NODE_ENV=production
```

anandamaryon1 (commented):
I can't pretend to understand how this works, but from a user point of view, it looks great. Tiny nitpick, feel free to disregard, but perhaps the html page title could be 'Enter password' instead of sign in (since the user won't have an account).

Perhaps @sarawilcox you can review the content and @roshaanbajwa can review the technical side?

Thanks for this contribution!

sarawilcox:
I think the password page is fine. Like @anandamaryon1, I think we should change the page title "Enter password - NHS prototype kit". (Note that we lowercase "prototype kit".)

sarawilcox:
Can I check about the update to this page, please, @frankieroberto ? https://nhsuk-prototype-kit.azurewebsites.net/docs/how-tos/publish-your-prototype-online

We've got this content currently. Does what you're suggesting apply to Railway as well as Heroku? Do we know what version of the NHS prototype kit will require a user name too? 

> To set a password, check your hosting services documentation on how to set "environment variables". (It may have a slightly different name like "config vars" or "variables".)
>
> Set environment variables. For example:
> 
> PROTOTYPE_USERNAMEnameofservice
> 
> PROTOTYPE_PASSWORDyourpassword123

> In Railway, you also need to create:
> 
> - a variable called NODE-ENV and set the value to production
> - a variable called USE_AUTH with the value of true

frankieroberto:
@sarawilcox @anandamaryon1 I've updated the page title, good catch.

I’ve also changed the password variable to use `PROTOTYPE_PASSWORD` instead of just `PASSWORD`, for consistency with how the kit is currently set up.

This also makes the documentation easier - I've updated this to describe how to set `PROTOTYPE_PASSWORD`, and added a second paragraph to say that you’ll also need to set `PROTOTYPE_USERNAME` if using an older version.

Let me know what you think!

@Tosin-Balogun, @edwardhorsford, @paulrobertlloyd or others: do you fancy giving this a quick test and review? It's been a few weeks since I worked on this so it could do with another pair of eyes...

docs/views/how-tos/publish-your-prototype-online.html
```diff
@@ -33,13 +33,13 @@ <h2>Hosting services</h2>
...
       
       <h2>Setting a password</h2>
-      <p>When running the prototype kit online, you must set a username and password. This is to stop anyone accidentally finding your prototype and mistaking it for a real service.</p>
+      <p>When running the prototype kit online, you must set a password. This is to stop anyone accidentally finding your prototype and mistaking it for a real service.</p>
       <p>To set a password, check your hosting services documentation on how to set "environment variables". (It may have a slightly different name like "config vars" or "variables".)</p>
-      <p>Set environment variables. For example:</p>
-      <p><code class="app-code">PROTOTYPE_USERNAME</code><code class="app-code">nameofservice</code></p> 
-      <p><code class="app-code">PROTOTYPE_PASSWORD</code><code class="app-code">yourpassword123</code></p>
+      <p>The password should be set using the variable <code class="app-code">PROTOTYPE_USERNAME</code>.</p>
```
anandamaryon1: Should this say PROTOTYPE_PASSWORD not USERNAME?

Tosin-Balogun:
> @Tosin-Balogun @sarawilcox @anandamaryon1 hiya. Sorry, I was on holiday so didn't see this update straight away.
> 
> I've deployed a quick demo of the password page here: https://nhsuk-prototype-password-page-b7626ee05c0c.herokuapp.com/ (password `testing`).
> 
> You can also test it locally, by adding these to a `.env` file:
> 
> ```
> PASSWORD=testing
> NODE_ENV=production
> ```

@frankieroberto This works well and is straight forward, thank you for the upgrade kind sir :) 

This change means:

- people might need to be prompted update their kit
- people need to be made aware of the change, perhaps at design huddle, show & tells, external facing slack, blog post
- guidance documentation update both on the website and the prototype kit template docs (@sarawilcox already doing)

docs/views/how-tos/publish-your-prototype-online.html
frankieroberto: Err, yes, opps. Thanks for spotting! Fixed now.

frankieroberto:
@Tosin-Balogun thanks, glad it worked for you! 😌  I borrowed heavily from the GOVUK prototype kit changes, but couldn’t copy and paste it all identically as there were some differences to the set up.

Thoughts on how best to document this change welcome! There already seems to be some guidance on how to do an update: https://nhsuk-prototype-kit.azurewebsites.net/docs/how-tos/updating-the-kit - which is a bit of a lengthy process. Perhaps this could be better signposted somehow though?

anandamaryon1 (approved):
Happy with this. Thanks for the work 🥇

roshaanbajwa (approved):

PR #356, opened 2024-09-19
Upgrade nhsuk frontend version to v9.0.1

frankieroberto:
Updates NHS.UK Frontend to version 9.0.1 plus associated changes.

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 :new: **New features**
 
-- Update to NHS.UK frontend v9.0.0 - see [version 9.0.0 release notes](https://github.com/nhsuk/nhsuk-frontend/releases/tag/v9.0.0) for details on the breaking changes
+- Update to NHS.UK frontend v9.0.1 - see [version 9.0.0 release notes](https://github.com/nhsuk/nhsuk-frontend/releases/tag/v9.0.0) for details on the breaking changes
 - Drop `outerContent` block and use `beforeContent` block for both breadcrumbs and back links
 
 ## 4.12.0 - 15 August 2024
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 :new: **New features**
 
-- Update to NHS.UK frontend v9.0.0
+- Update to NHS.UK frontend v9.0.0 - see [version 9.0.0 release notes](https://github.com/nhsuk/nhsuk-frontend/releases/tag/v9.0.0) for details on the breaking changes
 - Drop `outerContent` block and use `beforeContent` block for both breadcrumbs and back links
 
 ## 4.12.0 - 15 August 2024
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # NHS.UK prototype kit Changelog
 
+## Unreleased
+
+:new: **New features**
+
+- Update to NHS.UK frontend v9.0.0
+- Drop `outerContent` block and use `beforeContent` block for both breadcrumbs and back links
+
 ## 4.12.0 - 15 August 2024
 
 :wrench:
--- a/app.js
+++ b/app.js
@@ -51,6 +51,7 @@ app.use(cookieParser());
 // Nunjucks configuration for application
 const appViews = [
   path.join(__dirname, 'app/views/'),
+  path.join(__dirname, 'node_modules/nhsuk-frontend/packages/macros'),
   path.join(__dirname, 'node_modules/nhsuk-frontend/packages/components'),
   path.join(__dirname, 'docs/views/'),
   path.join(__dirname, 'lib/prototype-admin/'),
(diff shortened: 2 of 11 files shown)


frankieroberto:
@anandamaryon1 @edwardhorsford @paulrobertlloyd think I've caught all the updates needed, but let me know if you spot anything I’ve missed!

anandamaryon1:
Thanks @frankieroberto 

Updated the Check answers page template to conform with the latest headings guidance.

anandamaryon1 (dismissed):
Looks good. Can't think of anything else that needs amending. (But will appreciate other's taking a look)

frankieroberto:
@anandamaryon1 thanks - I didn’t check all the example and docs pages.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # NHS.UK prototype kit Changelog
 
+## Unreleased
+
+:new: **New features**
+
+- Update to NHS.UK frontend v9.0.0
+- Drop `outerContent` block and use `beforeContent` block for both breadcrumbs and back links
```
edwardhorsford, line 8: This could possibly have more narrative as users may need to update their layouts?

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # NHS.UK prototype kit Changelog
 
+## Unreleased
+
+:new: **New features**
+
+- Update to NHS.UK frontend v9.0.0
```
edwardhorsford: Add link to the release?

CHANGELOG.md
edwardhorsford: Or else link to the release notes in NHS.UK frontend that tell them what to do.

edwardhorsford (changes requested):
Looking good. Have run locally and it seems to work ok.

I think the changelog could be more extensive as this is probably a breaking change - we'll need to release v9 of the prototype kit.

CHANGELOG.md
frankieroberto: Good idea. Although we might also want to put that narrative on the [update page](https://nhsuk-prototype-kit.azurewebsites.net/docs/how-tos/updating-the-kit) as most users doing upgrades won't see the changelog by default?

anandamaryon1: Agree. Perhaps a more general mention, something like: check the frontend release notes for breaking changes that you may need to incorporate? To save having version specific stuff on the update page.

app.js
```diff
@@ -51,6 +51,7 @@ app.use(cookieParser());
 // Nunjucks configuration for application
 const appViews = [
   path.join(__dirname, 'app/views/'),
+  path.join(__dirname, 'node_modules/nhsuk-frontend/packages/macros'),
```
edwardhorsford: I think this should probably be sorted alphabetically.

docs/views/templates/blank-transactional.html
```diff
@@ -9,11 +9,10 @@
 {% block beforeContent %}
 {% endblock %}
```
edwardhorsford: Remove these

CHANGELOG.md
frankieroberto: Have added a link and mention in 3baa3468081e9d0a0dfd9bd080a98cf2e6cfaedb - is that enough?

CHANGELOG.md
frankieroberto: done in 3baa3468081e9d0a0dfd9bd080a98cf2e6cfaedb

docs/views/templates/blank-transactional.html
frankieroberto: Opps, well spotted!

frankieroberto: Fixed in 2babbad60a0aa35d2a5e085cf72280487025622e

app.js
frankieroberto: Done in 2e2d2150ece1b7fc9d17508680f93c3b81455f0f

docs/views/templates/check-your-answers.html
```diff
@@ -6,15 +6,22 @@
...
+    "text": "Back"
+  }) }}
+{% endblock %}
+
 {% block content %}
   <div class="nhsuk-grid-row">
     <div class="nhsuk-grid-column-two-thirds">
 
-      <h1 class="nhsuk-u-margin-top-4">
+      <h1 class="nhsuk-heading-l">
```
edwardhorsford, line 20: I wonder if we should redefine `h1`, `h2`, and `h2` in scss so that these don't need to be set on every heading.

The pro would be that prototypes get the right defaults for transactional services - and reduce the risk of designs accidentally using the wrong one. The con would be that they _might_ be surprised a not needing a class when working on these yet on production the font size is wrong.

frankieroberto: @edwardhorsford one for the backlog? I'm not so sure about it - kinda think we should be as similar to production as possible (often catches people out with the govuk prototype kit)

edwardhorsford: Ok, let's leave for now - but consider how we can improve this in the future.

edwardhorsford (dismissed):

CHANGELOG.md
edwardhorsford: I think it should be more explicit that users will need to make changes to templates for the back links.

edwardhorsford (approved):

PR #385, opened 2024-11-05
Remove duplicate guides from kit

### LINKED ISSUE #384: Remove docs from being embedded in prototype kit itself
The documentation (install guide, how-tos, etc) have been moved into a separate repo and website: http://github.com/nhsuk/nhsuk.service-manual.prototype-kit.docs

These can now be removed from the kit itself, to avoid the duplication and ensure that users always have access to the most up-to-date documentation.

frankieroberto:
This removes the duplicate install guide, how to guides, branching examples and passing data from page to page examples, and the documentation homepage from kit.

These guides are all now hosted separately in on the [NHS.UK prototype kit](https://prototype-kit.service-manual.nhs.uk) website and maintained within a separate repository: [nhsuk.service-manual.prototype-kit.docs](http://github.com/nhsuk/nhsuk.service-manual.prototype-kit.docs).

This is intended to allow us to update the guidance more frequently (outside of the prototype kit release cycle). It also means that uses of the kit will always access the most up-to-date versions of the guidance, even if their prototype is a few versions old. 

The one acknowledged downside is that it means you now need to be online to view the guidance, vs having a local copy.

The one bit I’ve not removed is the collection of example page templates. These are also duplicated on the prototype kit website at https://prototype-kit.service-manual.nhs.uk/page-templates - however they might be more useful to have locally, as it’ll be easier to copy-and-paste them.  I've created a separate issue to consider what the best thing to do with the templates is: [#60](https://github.com/nhsuk/nhsuk.service-manual.prototype-kit.docs/issues/60).

Fixes #384.

## Checklist

- [x] CHANGELOG entry


anandamaryon1:
Agree that it's best to leave in the page templates for now, as you say, it's helpful to have them 'to hand' when working in the kit.

anandamaryon1 (approved):

PR #386, opened 2024-11-05
Remove GitPod config

frankieroberto:
[GitPod](https://www.gitpod.io) looks kinda interesting - a cloud-based development environment I think?

I don’t know if anyone’s still using this for this project though? And perhaps GitHub Codespaces would do the same thing, but within the NHS's existing GitHub account billing?

Suggest we remove the config and documentation from the README for now, unless anyone knows a good reason to keep it?


frankieroberto:
Gitpod still works and is being used by @sarawilcox, @anandamaryon1 and team, so I'll close this for now. It sounds like it could be a useful tool to potentially point others to too, although we should also investigate how it compares to GitHub Codespaces and whether there’s any overlap or if one is better than the other? (I’m not super familiar with either) 

See also: https://github.com/nhsuk/nhsuk.service-manual.prototype-kit.docs/issues/70

PR #395, opened 2024-11-11
Move template to main app folder

frankieroberto:
This is another step towards being able to remove the `docs/` folder.

The `template.html` file is currently inherited by all pages within the main `app/` folder, but confusingly is located within `docs/`.


anandamaryon1 (approved):
@frankieroberto Makes sense. 

Out of interest, the template seems to import all (or most of) the components. Assume this is just so that users don't have to add the import in themselves each time they add a component to their page? If so, I think we need to add the new task list (and make a note to always add to this list when new components are released)

frankieroberto:
Hmm having chatted to @edwardhorsford we think it’s best to leave this in `docs/` for now, to make future updates to it easier.

Possibly it could move to `lib/` rather than `docs/`.

Longer-term it would be better to have the page template be part of `nhuk-frontend` instead, and then import the template from there (this is what `govuk-frontend` does). I’ll create a separate issue in the backlog for that.

frankieroberto:
Follow-up issue on NHS Frontend: https://github.com/nhsuk/nhsuk-frontend/issues/1072

PR #396, opened 2024-11-11
Remove Google page template and assets

frankieroberto:
I can see why this might be helpful, if you’re mocking up a journey with starts with a search engine, but including it in the kit, complete with the logos, feels a bit risky copyright-wise, and also hard to maintain?

It also doesn’t currently include the search results or search-as-you-type feature, so I’m not sure how useful it is?


vickytnz (approved):
Agree that this can be removed. My only concern was whether this is being used by teams, however after searching for 'google.html' on the Gibhub orgs [nhsuk](https://github.com/nhsuk/
), [NHSDigital](https://github.com/NHSDigital/) (including internal) and the [NHSBSA](https://github.com/nhsbsa), the only example I can find of actual use rather than a template is[ find a gp 3 years ago](https://github.com/NHSDigital/prototype-gp-registration/blob/a6a6331279061d6c8689c5c08f77035a47121892/app/views/user-research/18-nov/find-gp/google.html#L4). This supports that it's not being used and won't be missed.

frankieroberto:
@vickytnz 👍  I think even if teams had used it, they'd likely have copied the template from `docs/` into `app/` which wouldn’t then get removed if following the update instructions anyway?

vickytnz:
Yeah for me it was even the 'if we remove it will anyone notice that it's gone?' and I think that from use the answer is 'no'

PR #397, opened 2024-11-11
Remove unused images

frankieroberto:
These images are used by the install guides, which are now in a separate repository: https://github.com/nhsuk/nhsuk.service-manual.prototype-kit.docs

Forgot to remove these as part of #385.


vickytnz (approved):
Docs are now ringfenced so ok to remove

PR #402, opened 2024-11-12
Log errors when session data defaults crashes

### LINKED ISSUE #401: Errors in session-data-defaults.js crash the kit, but the error source doesn't show in console
If the kit crashes because of an error in session data defaults it'll say so in the console, but it won't show a helpful error message - eg indicate the line.

I believe the GOV.UK prototype kit used to have this bug and it was fixed at some point.

edwardhorsford:
## Description
Fixes #401 

[Fix copied from GOV.UK](https://github.com/alphagov/govuk-prototype-kit/pull/1050) (where I amusingly raised the same issue 5 years ago)

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -250,6 +250,7 @@ try {
   sessionDataDefaults = require(sessionDataDefaultsFile);
 } catch (e) {
   console.error('Could not load the session data defaults from app/data/session-data-defaults.js. Might be a syntax error?'); // eslint-disable-line no-console
+  console.error(e)
 }
 
 // Middleware - store any data sent in session, and pass it to all views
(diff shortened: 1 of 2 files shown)


lib/utils.js
```diff
@@ -250,6 +250,7 @@ try {
   sessionDataDefaults = require(sessionDataDefaultsFile);
 } catch (e) {
   console.error('Could not load the session data defaults from app/data/session-data-defaults.js. Might be a syntax error?'); // eslint-disable-line no-console
+  console.error(e)
```
frankieroberto: Rename `e` to `error`?

frankieroberto (dismissed):

lib/utils.js
edwardhorsford: Fixed.

frankieroberto (approved):

PR #409, opened 2024-11-13
Move example templates into lib folder (and drop NHS website templates)

### LINKED ISSUE #152: Remove jQuery
# Bug Report

## What is the issue?

It looks like the kit contains a fixed copy of jQuery v3.3.1 minimised - surely this is a bad idea? 

The current version is 3.5.1 and, when developing, I would want the non-minimised version as well for debugging.

It would surely be better to include jQuery as a dependency in package.json and configure webpack accordingly?

Download the current zip from the website.

Any.

N/A

frankieroberto:
Following discussion in nhsuk/nhsuk-prototype-kit-package#60 and https://github.com/nhsuk/nhsuk.service-manual.prototype-kit.docs/issues/60 this moves the example templates in `/lib/` and also drops the NHS.UK website templates (which remain on https://prototype-kit.service-manual.nhs.uk/page-templates for now).

This is partly as they’re no longer documentation but purely example page templates, and partly to try and make future updating easier by moving towards having a single folder (currently called `lib`) which users could manually update by copying and pasting (or possibly using a script or in future an npm package).

I’ve named the new folder with the templates `lib/example-templates`.

Because the NHS.UK website templates have been dropped, this also means that the `nhs.scss` can be deleted, as can the images used in the NHS website templates, and also jQuery (which resolves nhsuk/nhsuk-prototype-kit#152), which was only used in those templates.

Have also updated and simplified the `app.js` file where I can, although perhaps we can go further.

## Checklist

- [x] CHANGELOG entry


paulrobertlloyd:
Hmmmm. Allow me to ask a more fundamental question: how are these examples used (or intended to be used)? Are they meant to be starting points for new pages? Or are they meant to be pages you can copy and paste wholesale to replicate a flow, say from the NHS homepage, to a section of the NHS website, to your service? (Am guessing both!)

Does it matter that these templates are likely out-of-date (certainly the case for the NHS homepage). Do these examples need to be kept up to date (urgh)?

There’s also something about having a folder of large binary image assets, checked into a repo, and that need to be copied (or possibly in the future, packaged up), which feels like a bit of an anti-pattern.

Is the prototype kit repo the right place for these examples? Is it another repo, or maybe with an alongside the documentation?

It’d of course be lovely if this kit worked like the GOV.UK one, and these templates were part of a workflow, but given that’s not the case right now, what to do. I’m not sure.

Lots of questions! Some of which need answering I think before moving things around.

frankieroberto:
@paulrobertlloyd those are all good questions that I’ve had too! Currently we’ve also now duplicated the templates so we have 2 copies of them to maintain, in the kit and in the docs! 

I’ve added https://github.com/nhsuk/nhsuk.service-manual.prototype-kit.docs/issues/60 as a place to discuss the general principle of if/where to add the templates.

I was thinking that moving them could be a more immediate stopgap whilst we make a longer-term decision?

frankieroberto:
@paulrobertlloyd @vickytnz @anandamaryon1 @edwardhorsford ok I've updated this PR with some additional changes. It now also deletes the NHS.UK website templates (which we’re leaving on the Prototype Kit website for now), and jQuery.

In future we could go further and also delete some of the remaining templates (eg Check your answers or Confirmation page) as they get added to the NHS design systems as patterns.

For now, merging this would mean that, of the templates which remain, they’re exclusively either only in the kit itself, or only on the website - so we at least resolve the duplication.

We’re also one step closer to having as much of the code as possible within the `/lib` folder for easier updates.

vickytnz (dismissed):
All looks good. I have fixed one broken link in the footer (forgot to PR, oops)

Only thing to think about (but not blocking) - the content examples aren't strictly the same as the NHS website. 

I wonder if those pages:
* should have the service title hidden (similar to how it's hidden on GOV.UK prototype kit pages for website stuff)
* maybe should have a bit on the index page to make it clear that they're simplified versions, potentially even with a note to get stuff from the design system or the team. Unless you think that the details on the actual pages are enough.

Use simplified versions of NHS.UK website pages for user research [list] If you need content pages that look exactly like the NHS.UK website, talk use the NHS Design System components or get help from the service manual team.

frankieroberto:
@vickytnz good point. The `blank-nhsuk.html` template seems so close to `content-page.html` I think it can be removed, so I’ve done that in 293bb9e.

Also noticed that the "Mini-hub" page is already documented in the Design system: https://service-manual.nhs.uk/design-system/patterns/mini-hub - so perhaps that could also be removed in future.

The only thing missing from the Nunjucks code in the Mini-hub page on the design system is `{% extends 'layout.html' %}` at the top, which you’d need within the prototype kit.

vickytnz (dismissed):

lib/example-templates/index.html
```diff
@@ -18,15 +18,14 @@
 
       <h1>Page templates</h1>
 
-      <p>You can find these templates in the <strong>/docs/views/templates</strong> folder.</p>
+      <p>You can find these templates in the <strong>/lib/example-templates</strong> folder.</p>
```
anandamaryon1, line 21: Maybe this should say `/lib/example-templates/templates` since they are actually in that folder, to avoid confusion?

anandamaryon1:
Some good wins in here, removing jQuery, image assets and changing the links to `#` 👍

lib/example-templates/index.html
frankieroberto: @anandamaryon1 good point. Maybe I should move those templates up a level into `/lib/example-templates`?

anandamaryon1: I think it'd be neater to keep them in their own folder, so they're not mixed in with the app pages eg. `index.html`, but happy either way, if you feel otherwise.

edwardhorsford (dismissed):
I don't think this should remove jQuery - or at least that should be a separate PR.

Doing so should probably be considered a breaking change - and it does not feel necessary or especially relevant for the purposes of this PR.

frankieroberto:
> I don't think this should remove jQuery - or at least that should be a separate PR.
> 
> Doing so should probably be considered a breaking change - and it does not feel necessary or especially relevant for the purposes of this PR.

@edwardhorsford Good point. Have restored in 16d4dd9. We can remove in a future breaking change release (given it’s no longer used)?

lib/example-templates/index.html
frankieroberto: @anandamaryon1 I've moved them up a level in 9729500. Feels simpler. Possibly the `layout.html` in `lib/example-templates` could be removed in future (it's now identical to `app/views/layout.html`.

vickytnz (approved):
This can go forward but also needs two more tickets:

1. the basic prototype guidance will need to be updated when this codes live as all of the links to templates (for example https://prototype-kit.service-manual.nhs.uk/how-tos/build-basic-prototype/open-prototype-in-editor ) will now be wrong (possibly with a note to older versions) 
2. there are still inconsistencies with the templates against the design system (for example 'back' instead of 'go back' - these need to be fixed and then also code and screenshots changed in the basic prototype tutorial

PR #410, opened 2024-11-13
Use strict https headers and redirect from http to https in production mode

### LINKED ISSUE #141: Force https when running in production
Most websites are now https and chrome gives you a warning if a website you are visiting isn't using http.

The prototype kit doesn't force this. So if you deploy to heroku you can have both a http and https version of the website. https is more secure.

In the app team we also noticed that when running the prototype kit as a PWA (to simulate the app) that if we used the http version a massive warning bar came up when a user used a field. This happened on every text input.

We should force (or allow the kit to force) https. More secure. Saves issues for people navigating and sharing the http version for testing and documentation.

frankieroberto:
This aims to avoid browser warnings if users accidentally visit a prototype in non-secure mode.

It does this by adding the [`Strict-Transport-Security` HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security) and `Content-Security-Policy` to `upgrade-insecure-requests`.

This will ensure that all subsequent requests will automatically use https, even if the `http://` prefix is typed into the URL bar.

Additionally, the kit will automatically respond with a redirect to the https url whenever insecure requests are made.

This only happens when [`NODE_ENV`](https://nodejs.org/en/learn/getting-started/nodejs-the-difference-between-development-and-production) is set to `production`, to avoid any issues when running the prototype over localhost

Once your prototype is running with the header set, you can also submit the domain to https://hstspreload.org/ (although this doesn't work for subdomains) which will ensure the browsers at it to their list of domains to always use https on without even having to visit it first.

Fixes #141

## Checklist

- [x] CHANGELOG entry


frankieroberto:
@edwardhorsford @vickytnz ok to approve this?

We should then add some guidance to suggest adding the `NODE_ENV=production` env variable in hosting environments (Heroku, etc) - although some of them might add it by default.

frankieroberto:
Setting NODE_ENV=production also does some performance things in Express: https://expressjs.com/th/advanced/best-practice-performance.html#set-node_env-to-production

vickytnz:
oh wow - i didn't realise that was the case. I know that I had to do this for railway but didn't think I needed to for heroku but that does have some performance benefits

vickytnz:
> @edwardhorsford @vickytnz ok to approve this?
> 
> We should then add some guidance to suggest adding the `NODE_ENV=production` env variable in hosting environments (Heroku, etc) - although some of them might add it by default.


Reviewing the guidance, we already say that railway users have to do this. https://prototype-kit.service-manual.nhs.uk/how-tos/publish-your-prototype-online

```
In Railway, you also need to create:

a variable called NODE-ENV and set the value to production
a variable called USE_AUTH with the value of true
```
Unless this breaks Heroku (which I don't think it does), this should be OK to go without changes. 

My only thought is that whether this makes enough of a difference to performance that we need to headline it as adding it in and saying (the prototype will run without it but it will take images slower to load and use more data) 

EDIT: read your notes from the top, you're saying that it's done by default by Heroku so actually the guidance is probably good enough.

frankieroberto:
@vickytnz yep, I believe that's the case, although we should probably test it. Can't remember if I added it manually or not.

frankieroberto:
@anandamaryon1 could you review?

app.js
```diff
@@ -189,6 +189,15 @@ exampleTemplatesApp.get(/^([^.]+)$/, (req, res, next) => {
 
 app.use('/prototype-admin', prototypeAdminRoutes);
 
+if (process.env.NODE_ENV === 'production') {
+  app.use((req, res, next) => {
+    // Set Strict-Transport-Security header to
+    // ensure that browsers only use HTTPS
+    res.setHeader('Strict-Transport-Security', 'max-age=31536000; preload');
```
colinrotherham: Can I check if [`includeSubDomains` is mandatory](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security#preload) when `preload` is used?

```suggestion
    res.setHeader('Strict-Transport-Security', 'max-age=31536000; includeSubDomains; preload');
```

lib/utils.js
```diff
@@ -110,21 +110,6 @@ exports.findAvailablePort = function (app, callback) {
 }
 */
 
-/*
-// Redirect HTTP requests to HTTPS
-exports.forceHttps = function (req, res, next) {
-  if (req.headers['x-forwarded-proto'] !== 'https') {
-    console.log('Redirecting request to https')
-    // 302 temporary - this is a feature that can be disabled
-    return res.redirect(302, 'https://' + req.get('Host') + req.url)
```
colinrotherham, line 119: >This aims to avoid browser warnings if users accidentally visit a prototype in non-secure mode.

From the PR description, was the `res.redirect()` not working as expected?

So we know why it was replaced with the HSTS header

app.js
frankieroberto: Ah good point!

lib/utils.js
frankieroberto: This whole function was commented out (and not referenced elsewhere), but I’m not sure why. I imagine that relying on the `req.headers['x-forwarded-proto']` didn't work for all hosting environments?

frankieroberto:
I’ve verified this works on Heroku.

For Railway, they do [their own redirecting of HTTP -> HTTPS](https://docs.railway.com/reference/public-networking#technical-specifications) at the proxy layer for all apps, so the code is unneeded, but doesn’t do any harm.

lib/middleware/production.js
```diff
@@ -0,0 +1,22 @@
+// This Express middleware function sets some
+// HTTP headers which should only be set in deployed
+// environments
+
+function production(req, res, next) {
+  // Set Strict-Transport-Security header to
+  // ensure that browsers only use HTTPS
+  res.setHeader('Strict-Transport-Security', 'max-age=31536000; includeSubDomains; preload');
```
colinrotherham, line 8: Did we decide these didn't work unless `herokuapp.com` added headers too?

```suggestion
  res.setHeader('Strict-Transport-Security', 'max-age=31536000');
```

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Use strict https in production mode - ([PR 410](https://github.com/nhsuk/nhsuk-prototype-kit/pull/410))
```
colinrotherham: Do we need a line about automatic redirects to https too?

lib/middleware/production.js
frankieroberto: We did, but they might be worth keeping in case you deploy elsewhere, or use a custom top-level domain on Heroku or elsewhere. (Unlikely but you never know).

I don’t think they’ll do any harm when used on a subdomain, they’ll just be ignored?

CHANGELOG.md
frankieroberto:
```suggestion
- Use strict https and automatic http to https redirects in production mode - ([PR 410](https://github.com/nhsuk/nhsuk-prototype-kit/pull/410))
```

frankieroberto: Good spot. Updated.

colinrotherham (approved):
Nice work, let's give it a go. Can always tweak some more

PR #423, opened 2024-11-19
Update default index page

frankieroberto:
Following #419, this updates the default index page to give clearer guidance on how to:

* set the service name (which is now set to "Service name goes here" by default)
* edit the index page
* use the external links and included example templates

Inspired by the GOV.UK Prototype kit default index page.

## Screenshots

| Before | After |
| -------|-------|
| index-page-before | index-page-after |

## Checklist

- [x] CHANGELOG entry


vickytnz:
I think Sara may need to review this, on the same page we have:

* NHS.UK prototype kit
* NHS Prototype kit

I *think* that both now should be NHS prototype kit (without the .UK and lower case) but don't really know right now.

vickytnz (dismissed):

frankieroberto:
@vickytnz thanks, have updated from `main` now with your changes to capitalisation and NHS.UK -> NHS.  Could you re-review?

vickytnz (approved):
All looks good - I've also fixed the broken link in the footer that we all missed when removing the example page 🙈

PR #440, opened 2024-12-10
Move middleware into lib folder

frankieroberto:
This moves the `middleware` folder into the `lib` folder and updates file references to them.

This is to enable easier updates to the kit, as the instructions could be simplified to downloading the new version and replacing the `lib` folder, plus a few manual changes to files outside `lib` (which we can hopefully further reduce).

Part of nhsuk/nhsuk-prototype-kit-package#60.


vickytnz (approved):
Downloaded, installed and ran this version and all seems to work.

PR #451, opened 2025-01-07
Removing image attribute from devcontainer to ensure standard linux image is used by Github Codespaces

paulmsmith:
## Description

Removing the attribute to set the image, as the default linux image is excluded from (does not contribute to) storage usage limits of a Codespace.

## Checklist

- [x] CHANGELOG entry


frankieroberto (approved):
Thanks!

PR #482, opened 2025-03-15
Add reset data feature

### LINKED ISSUE #389: Add a feature to clear session data
## What is the issue?

There is no way in the prototype kit to clear data. This is possible in the GOV.UK prototype kit with pages designed to do it. https://prototype-kit.service.gov.uk/docs/pass-data

frankieroberto:
This adds a page which will let users 'reset' any data entered in the session back to whatever defaults are set in `data/session-data-defaults.js` (or an empty `{}` object if that is blank).

Fixes #389

Inspired by the equivalent feature in the GOV.UK Prototype Kit.


paulmsmith (commented):
Might be nice for a success page for the reset instead of just redirecting to '/'? Guess it depends on how often somebody will perform the reset?

paulmsmith (dismissed):
A thumbs up from me.

joelanman:
I think a good default is to return to the page you were on. This might work on the clear data page itself:

```
response.locals.returnURL = request.get('Referrer') || '/'
```

edwardhorsford:
I prefer returning to the page you came from too. Having to visit a page, press a button, then navigate back feels annoying. I'd prefer a link in the footer.

In my prototype I've set it up to [perform the clear from the page you're on via ajax](https://github.com/NHSDigital/manage-screening-events-prototype/blob/main/app/assets/javascript/main.js#L49-L77).

Requires a small change in the route to accept:
```
if (req.headers['accept'] === 'application/json') {
      return res.json({ success: true })
    }
```

lib/middleware/prototype-admin-routes.js
```diff
@@ -34,4 +34,13 @@ router.post('/password', (req, res) => {
   }
 });
 
+router.get('/reset', (req, res) => {
+  res.render('reset');
+});
```
edwardhorsford, line 39: Is this route necessary? shouldn't this get auto-routed anyway?

lib/middleware/prototype-admin-routes.js
```diff
@@ -34,4 +34,13 @@ router.post('/password', (req, res) => {
...
+
+router.post('/reset', (req, res) => {
+  req.session.data = {};
+  res.redirect('/');
+});
```
edwardhorsford:
```suggestion
router.all('/reset-session-data', (req, res) => {
  req.session.data = {};
  res.redirect('/');
});
```

Would let this work via GET too so someone could put it in the footer. And then change the url so as to not conflict with the admin page.

frankieroberto:
@edwardhorsford @paulmsmith I'm away all next week, but happy for anyone to pick this up if you fancy it.

Redirecting back sounds good.

frankieroberto:
@edwardhorsford @paulmsmith @colinrotherham @joelanman I’ve picked this up again and I’ve added [c69be10](https://github.com/nhsuk/nhsuk-prototype-kit/pull/482/commits/c69be10a9969e452c05d9f797e19cc30cf28183c) which:

1. sets the 'Back' link to the previous page (based on Referrer header, defaulting to `/`)
2. sets the previous page as a hidden value in the form to reset data
3. adds a confirmation page that data has been reset, which includes a link to the page you were on before (defaulting to `/` again)

Would you mind testing it?

lib/middleware/prototype-admin-routes.js
```diff
@@ -34,4 +34,20 @@ router.post('/password', (req, res) => {
   }
 });
 
+router.get('/reset', (req, res) => {
+
+  const backHref = req.get('Referrer') || '/'
```
frankieroberto: Probably need to find a way to sanitize the Referrer header to check it’s a valid local path, given that it’s set by the user/browser.

lib/prototype-admin/reset.html
```diff
@@ -0,0 +1,30 @@
...
+{% block pageTitle %}
+  Reset data - NHS prototype kit
+{% endblock %}
+
+{% block beforeContent %}
+  {{ backLink({
+      "text": "Back",
+      "href": backHref
+    })
+  }}
```
colinrotherham: Minor indenting tweak here

```suggestion
  {{ backLink({
    "text": "Back",
    "href": backHref
  }) }}
```

lib/prototype-admin/reset.html
```diff
@@ -0,0 +1,30 @@
...
+    <div class="nhsuk-grid-column-two-thirds">
+      <h1 class="nhsuk-heading-l">Reset data</h1>
+
+      <p class="nhsuk-body">This will reset all of the data entered in this session.</p>
+
+      <form action="/prototype-admin/reset-session-data" method="post">
+        <input type="hidden" name="previousPath" value="{{ backHref }}">
+        {{ button({
+          text: "Reset all data"
+        })}}
```
colinrotherham:
```suggestion
        }) }}
```

lib/middleware/prototype-admin-routes.js
```diff
@@ -34,4 +34,20 @@ router.post('/password', (req, res) => {
...
+  res.render('reset', {
+    backHref
+  });
+});
+
+router.all('/reset-session-data', (req, res) => {
+  req.session.data = {};
+  const returnPath = req.body.previousPath || '/'
```
colinrotherham: We'll need to make sure we're not redirecting to untrusted referrers
https://expressjs.com/en/advanced/best-practice-security.html#prevent-open-redirects

But since the `returnPath` name is so good, we could do it all with that?

1. Visit prototype page `/example`
2. Click link **Reset data** to GET `/prototype-admin/reset?returnPath=%2Fexample`
3. Click button **Reset all data** to POST `/prototype-admin/reset?returnPath=%2Fexample`
4. Redirect to `returnPath` (sanitised) from `req.query`

lib/middleware/prototype-admin-routes.js
```diff
@@ -34,4 +34,20 @@ router.post('/password', (req, res) => {
...
+  res.render('reset-done', {
+    returnPath
+  });
```
colinrotherham: The problem with `res.render()` here is we're showing it after a POST

The watch task will reload the page (back to a GET) as soon as you edit a stylesheet etc

Can we redirect to a stable success page instead?

```suggestion
  const successPath = '/prototype-admin/reset-session-data-success'
  res.redirect(`${successPath}?returnPath=${returnPath}`);
```

lib/prototype-admin/reset-done.html
```diff
@@ -0,0 +1,17 @@
...
+{% endblock %}
+
+{% block content %}
+  <div class="nhsuk-grid-row">
+    <div class="nhsuk-grid-column-two-thirds">
+      <h1 class="nhsuk-heading-l">Data reset</h1>
+
+      <p class="nhsuk-body">The session data has been reset.</p>
+
+      <p><a href="{{ returnPath }}" class="nhsuk-link">Return</a>
```
colinrotherham: Does this one need `nhsuk-body` like the paragraph above?

```suggestion
      <p class="nhsuk-body"><a href="{{ returnPath }}" class="nhsuk-link">Return</a>
```

lib/prototype-admin/reset.html
```diff
@@ -0,0 +1,30 @@
...
+
+{% block content %}
+  <div class="nhsuk-grid-row">
+    <div class="nhsuk-grid-column-two-thirds">
+      <h1 class="nhsuk-heading-l">Reset data</h1>
+
+      <p class="nhsuk-body">This will reset all of the data entered in this session.</p>
+
+      <form action="/prototype-admin/reset-session-data" method="post">
+        <input type="hidden" name="previousPath" value="{{ backHref }}">
```
colinrotherham: Regarding the single `returnPath` idea

Here's how you POST but with `req.query` still containing `returnPath`

```suggestion
      <form action="/prototype-admin/reset-session-data?returnPath={{ returnPath }}" method="post">
```

lib/middleware/prototype-admin-routes.js
colinrotherham: Nice comment, didn't see it in time but same point in https://github.com/nhsuk/nhsuk-prototype-kit/pull/482#discussion_r2056384301 👀

lib/middleware/prototype-admin-routes.js
frankieroberto: @colinrotherham It’s linked to from every page (in the footer), so there would need to be some way to dynamically add the current path to the link - I'm not sure that views have access to `req.path` by default, but maybe it could be set in some middleware?

lib/middleware/prototype-admin-routes.js
frankieroberto: I think the issue with that would be that you could then visit that stable success page using a regular GET without having actually cleared the data first, which would be misleading.

I’d usually use the POST->redirect->GET pattern, but maybe this is an exception? The GOVUK prototype kit renders the success page from the POST too... 🤔

lib/prototype-admin/reset-done.html
frankieroberto: I think the style is applied by default to `<p>`, and same with `nhsuk-link` (which actually does nothing at all), but no harm in adding them in case this changes in future.

lib/prototype-admin/reset.html
frankieroberto: I always forget that you can do a POST with a query string! Feels wrong somehow, even though it works just fine.

lib/middleware/prototype-admin-routes.js
colinrotherham: Shall we take a look tomorrow?

edwardhorsford: I pass [req.path to views via locals](https://github.com/NHSDigital/manage-breast-screening-prototype/blob/main/app/locals.js#L7) in my prototype.

lib/prototype-admin/reset.html
frankieroberto: I've kept hidden field for now - but could revert to query string if needed. (Not sure it matters either way)

lib/middleware/prototype-admin-routes.js
frankieroberto: Fixed now by not using Referrer and instead passing in the current page as a query string.

CHANGELOG.md
```diff
@@ -36,6 +36,8 @@ to
 
 - Remove ‘Check your answers’ example template, as this is now available on the NHS design system website ([PR 503](https://github.com/nhsuk/nhsuk-prototype-kit/pull/503))
 - Remove Confirmation page example template, as this is now available on the NHS design system website as a pattern ([PR 504](https://github.com/nhsuk/nhsuk-prototype-kit/pull/504))
+- Added a way to reset any data entered back to the default - ([PR 482](https://github.com/nhsuk/nhsuk-prototype-kit/pull/482))
+
```
colinrotherham, line 57: Extra new line can be chopped

```suggestion
```

app/locals.js
```diff
@@ -1,5 +1,6 @@
 module.exports = (config) => (req, res, next) => {
   res.locals.serviceName = config.serviceName;
+  res.locals.currentPage = req.originalUrl;
```
colinrotherham: Can we strip the full URL scheme `https://` and hostname?

Mainly because we're defaulting to `'/'` (path not full URL) in the admin routes

```js
const { pathname, search } = req.originalUrl

res.locals.serviceName = config.serviceName;
res.locals.currentPage = `${pathname}${search}`;
```

☝️ Then we only get `/path-to-page?optional=true`

lib/middleware/prototype-admin-routes.js
```diff
@@ -34,4 +34,19 @@ router.post('/password', (req, res) => {
...
+router.get('/reset', (req, res) => {
+  const returnPage = req.query.returnPage || '/';
+  res.render('reset', {
+    returnPage,
+  });
+});
+
+router.all('/reset-session-data', (req, res) => {
+  req.session.data = {};
+  const returnPage = req.body.returnPage || '/';
```
colinrotherham: Be aware we still need to reject `returnPage` if it's an external URL

Maybe something crude like checking it begins with `/` is enough?

```suggestion
  const returnPage = req.body.returnPage?.startsWith('/')
    ? req.body.returnPage // Local paths only
    : '/';

  req.session.data = {};

```

colinrotherham: But equally it might not be a string if someone had hand-crafted the POST (or hidden input)

app/locals.js
frankieroberto: `req.originalUrl` already only returns path and query string: https://expressjs.com/en/api.html#req.originalUrl

lib/middleware/prototype-admin-routes.js
frankieroberto: committed your suggestion. Feels good enough? Not sure how it could be a non-string?

app/locals.js
colinrotherham: My fault assuming `Url` meant URL, sorry about that!

lib/middleware/prototype-admin-routes.js
colinrotherham: We set `body-parser` `extended: true` in https://github.com/nhsuk/nhsuk-prototype-kit/blob/e60c9845adfe942076d4e50f530ac28031600a86/app.js#L105-L107

Which from memory means non-strings come from:

### POST as array value

```patch
- <input type="hidden" name="returnPage" value="{{ returnPage }}">
+ <input type="hidden" name="returnPage[0]" value="{{ returnPage }}">
```

### POST as object value

```patch
- <input type="hidden" name="returnPage" value="{{ returnPage }}">
+ <input type="hidden" name="returnPage[name]" value="{{ returnPage }}">
```

colinrotherham (approved):

PR #489, opened 2025-03-24
Downgrade sass to 1.77.5

frankieroberto:
This matches the version used by NHS Frontend itself, which was downgraded in https://github.com/nhsuk/nhsuk-frontend/pull/1064

The main reason for the downgrade is to temporary avoid some of the deprecation warnings introduced in later versions.

Probably we should pin the version of sass used in the prototype kit to the same version as used by NHS Frontend?

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -5,7 +5,8 @@ const gulp = require('gulp');
 const babel = require('gulp-babel');
 const browserSync = require('browser-sync');
 const clean = require('gulp-clean');
-const sass = require('gulp-sass')(require('sass'));
+const gulpSass = require('gulp-sass')
+const dartSass = require('sass-embedded')
 const nodemon = require('gulp-nodemon');
 
 // Local dependencies
@@ -19,7 +20,7 @@ function cleanPublic() {
   return gulp.src('public', { allowEmpty: true }).pipe(clean());
 }
 
-sass.compiler = require('sass');
+const sass = gulpSass(dartSass)
 
 // Compile SASS to CSS
 function compileStyles() {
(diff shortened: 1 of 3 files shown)


colinrotherham (approved):
Looks good

Shall we aim high and help @paulmsmith too?

I've not tried this but it'll be something like:

1. Silence dependencies with `quietDeps`
2. Add load path `.` to include `./node_modules` as dependencies
3. Add `silenceDeprecations: ['import']` to continue using `@import`

Maybe then we don't need to downgrade?

```patch
- function compileStyles() {
+ function compileStyles(done) {
    return gulp
      .src(['app/assets/sass/**/*.scss'])
-     .pipe(sass())
+     .pipe(
+       sass({
+         loadPaths: ['.'],
+         quietDeps: true,
+         silenceDeprecations: ['import']
+       }).on('error', done)
+     )
      .pipe(gulp.dest('public/css'))
-     .on('error', (err) => {
-       console.log(err);
-       process.exit(1);
-     });
  }
```

☝️ Was that `.on('error)` in the wrong place before?

frankieroberto:
@colinrotherham after downgrading to 1.77.5 I’m not seeing any sass deprecation warnings at all, so maybe silencing deprecations isn’t needed for now?

colinrotherham:
> @colinrotherham after downgrading to 1.77.5 I’m not seeing any sass deprecation warnings at all, so maybe silencing deprecations isn’t needed for now?

I know a lot of people like to update their packages?

We're going to see `@import` warnings from Sass 1.80.0 until 3.0.0 lands

Approved it already though, you decide!

PR #499, opened 2025-04-22
Use template from NHS Frontend

frankieroberto:
This switches the prototype kit to use the `template.njk` which is now included within NHS Frontend, rather than supplying its own, which was confusingly located within `lib/example-templates`.

## Checklist

- [x] CHANGELOG entry


colinrotherham (dismissed):

edwardhorsford (commented):
Should this be considered a breaking change? Possibly?

What do you think about putting the `imports` in their own file that sits outside of `app`? and then including that in `layout.html`.

app/views/layout.html
```diff
@@ -1,14 +1,50 @@
...
+{%- from 'select/macro.njk' import select %}
+{%- from 'skip-link/macro.njk' import skipLink %}
+{%- from 'skip-link/macro.njk' import skipLink %}
+{%- from 'summary-list/macro.njk' import summaryList %}
+{%- from 'tables/macro.njk' import table %}
+{%- from 'tabs/macro.njk' import tabs %}
+{%- from 'tag/macro.njk' import tag %}
+{%- from 'task-list/macro.njk' import taskList %}
+{%- from 'textarea/macro.njk' import textarea %}
+{%- from 'warning-callout/macro.njk' import warningCallout %}
```
frankieroberto: These lines aren't strictly needed, as you could import the relevant components on each page (and the service manual website includes these within its code snippets), but they make it easier to use the components without having to remember to import things - also keeps the code neater.

app/views/layout.html
```diff
@@ -1,14 +1,50 @@
...
+
 <!--
   This is the main layout where you can:
     - change the header and footer
     - add custom CSS and JavaScript
 -->
 
 <!-- Extends the layout from /lib/example-templates/template.html -->
```
anandamaryon1: > `<!-- Extends the layout from /lib/example-templates/template.html -->`

Does this comment need changing? Now that it's instead coming from nhsuk-frontend?

frankieroberto: err yes, good catch!

app/views/layout.html
edwardhorsford: I think we could consider moving these to their own include kept in `lib` or similar.

anandamaryon1 (dismissed):

app/views/layout.html
frankieroberto: Good idea.

frankieroberto:
@edwardhorsford let’s count this as a breaking change, since the `headCSS` will no longer be defined, and the layout files have changed. 👍

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # NHS prototype kit Changelog
 
+## Unreleased
+
+- Update NHS Frontend to 9.4.1
```
colinrotherham: Do we usually prefer ~NHS Frontend~ NHS.UK frontend?

```suggestion
- Update NHS.UK frontend to 9.4.1
```

app/views/layout.html
```diff
@@ -4,11 +4,11 @@
     - add custom CSS and JavaScript
 -->
 
-<!-- Extends the layout from /lib/example-templates/template.html -->
-{% extends "template.html" %}
+<!-- Extends the prototype kit template which in turn extends the NHS Frontend template -->
```
colinrotherham:
```suggestion
<!-- Extends the prototype kit template which in turn extends the NHS.UK frontend template -->
```

CHANGELOG.md
frankieroberto: Ooh good point, I keep forgetting this.

colinrotherham (approved):
Do we need a "Breaking changes" etc in the changelog?

frankieroberto:
@colinrotherham will add as part of a future PR. 👍

Technically given this repo is now a 'template repository', if you start out by using the repo as a template instead of the downloadable zip files, it’s kinda unversioned/evergreen.

PR #527, opened 2025-05-22
Preserve query when redirecting from POSTs

edwardhorsford:
## Description
This PR replicates a [similar PR I did on the GOV.UK prototype kit](https://github.com/alphagov/govuk-prototype-kit/pull/1120) 4 years ago.


colinrotherham (approved):
Looks good

The documentation for [`url.format()`](https://nodejs.org/api/url.html#urlformaturlobject) says:

>[Stability: 3](https://nodejs.org/api/documentation.html#stability-index) - Legacy: Use the WHATWG URL API instead.

But appreciate it's tried and tested already

edwardhorsford:
Chatting with @colinrotherham we think the revised code would be:
```js
app.post(/^\/([^.]+)$/, (req, res) => {
  const pathname = `/${req.params[0]}`;
  const searchParams = new URLSearchParams(req.query);
  res.redirect(searchParams.size ? `${pathname}?${searchParams}` : pathname);
});
```

This feels harder to read / parse and so I'm tempted to leave the code as is.

PR #531, opened 2025-05-23
Mark kit as compatible with node 22

frankieroberto:
This avoids warnings if users have Node 22 installed.

## Checklist

- [x] CHANGELOG entry


package.json
```diff
@@ -48,6 +48,6 @@
     "jest": "^29.7.0"
   },
   "engines": {
-    "node": "^20"
+    "node": "^20 || ^22"
```
colinrotherham: I'd recommend only LTS versions if you can

```suggestion
    "node": "^20.9.0 || ^22.11.0"
```

colinrotherham: Can you replace `20` with `ltd/jod` inside [**.nvmrc**](https://github.com/nhsuk/nhsuk-prototype-kit/blob/main/.nvmrc) too?

frankieroberto: @colinrotherham done

colinrotherham (approved):

PR #541, opened 2025-06-10
Make it easier to set title tags

### LINKED ISSUE #461: Make it easier to set page title with a common suffix
Currently we have this in the default layout file:

```
 <title>{% block pageTitle %}NHS prototype kit{% endblock %}</title>
```

This makes it possible to override the `<title>` tag on pages by setting the `pageTitle` block, but does not make it easy to follow the [best practice](https://service-manual.nhs.uk/accessibility/content) of having a common suffix for the site name, for example "Achalasia – NHS".

We could improve this by changing the layout to something like:

```
 <title>{% block pageTitle %}{% if pageName %}{{ pageName }} - {% endif %}}{{ serviceName }}{% endblock %}</title>
```

Which would then allow you to set a `pageName` attribute on pages and have your serviceName as a suffix.

You could still override the whole thing by using the existing `pageTitle` block.

frankieroberto:
This aims to make it easier to set page `<title>s` in accordance with [the guidance in the Service manual](https://service-manual.nhs.uk/design-system/styles/page-template#page-title).

## How it works

Instead of doing this:

```njk
{% block pageTitle %}
  Your address - Register for a GP - NHS
{% end block %}
```

you can now do this:

```
{% set pageName = "Your address %}
```

...and `- {{ serviceName }} - NHS` will be appended automatically.

In addition, you can use another variable to indicate that validation errors are present on the page:

```
{% set errorsPresent = true %}
{% set pageName = "Your address %}
```

...which will generate `Error: Your address - {{ serviceName }} - NHS`.

If a page does not need a separate name from the service name (for example, on the homepage for the service), then `pageName` can be set to `false` or omitted:

```
{% set pageName = false %}
```

...which will generate `{{ serviceName }} - NHS`

This logic can be override by using the `pageTitle` block as before.

Resolves #461. 

See also [Make it easier to set page title](https://github.com/alphagov/govuk-prototype-kit/pull/2030) in the GOV.UK Prototype Kit (resolving [a similar issue](https://github.com/alphagov/govuk-prototype-kit/issues/1975)), which uses the approach (but doesn’t include the `Error: ` prefix).

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/app/views/index.html
+++ b/app/views/index.html
@@ -10,10 +10,14 @@
     - add custom CSS and JavaScript
 -->
 
-<!-- Set the page title -->
-{% block pageTitle %}
-  {{ serviceName }}
-{% endblock %}
+<!--
+  Use this to set the page name which will appear at the
+  start of the <title> tag, followed by the service name.
+  This can be set to false on a homepage, where only the
+  service name is needed. You can also reuse this variable
+  within the <h1>, where they should be the same.
+-->
+{% set pageName = false %}
 
 <!-- For adding a breadcrumb or back link -->
 <!-- Code examples can be found at https://service-manual.nhs.uk/design-system/components/breadcrumbs and https://service-manual.nhs.uk/design-system/components/back-link -->
--- a/app/views/index.html
+++ b/app/views/index.html
@@ -13,9 +13,9 @@
 <!--
   Use this to set the page name which will appear at the
   start of the <title> tag, followed by the service name.
-  This can be set to false on a homepage, where only the
+  This is often not needed on the homepage, where only the
   service name is needed. You can also reuse this variable
-  within the <h1>, where they should be the same.
+  within the <h1>, where they are the same.
 -->
 {% set pageName = "" %}
 
--- a/app/views/index.html
+++ b/app/views/index.html
@@ -17,7 +17,7 @@
   service name is needed. You can also reuse this variable
   within the <h1>, where they should be the same.
 -->
-{% set pageName = false %}
+{% set pageName = "" %}
 
 <!-- For adding a breadcrumb or back link -->
 <!-- Code examples can be found at https://service-manual.nhs.uk/design-system/components/breadcrumbs and https://service-manual.nhs.uk/design-system/components/back-link -->
--- a/lib/templates/prototype-kit-template.njk
+++ b/lib/templates/prototype-kit-template.njk
@@ -35,4 +35,6 @@
 {%- from 'textarea/macro.njk' import textarea %}
 {%- from 'warning-callout/macro.njk' import warningCallout %}
 
+{% block pageTitle %}{% if errorsPresent %}Error: {% endif %}{% if pageName %}{{ pageName }} - {% endif %}{{ serviceName }} - NHS{% endblock %}
+
 {% set assetPath = '/nhsuk-frontend/assets' %}
--- a/lib/templates/prototype-kit-template.njk
+++ b/lib/templates/prototype-kit-template.njk
@@ -35,6 +35,6 @@
 {%- from 'textarea/macro.njk' import textarea %}
 {%- from 'warning-callout/macro.njk' import warningCallout %}
 
-{% block pageTitle %}{% if errorsPresent %}Error: {% endif %}{% if pageName %}{{ pageName }} - {% endif %}{{ serviceName }} - NHS{% endblock %}
+{% block pageTitle %}{% if errorsPresent %}Error: {% endif %}{% if pageName %}{{ pageName }} – {% endif %}{{ serviceName }} – NHS{% endblock %}
 
 {% set assetPath = '/nhsuk-frontend/assets' %}
--- a/lib/templates/prototype-kit-template.njk
+++ b/lib/templates/prototype-kit-template.njk
@@ -35,6 +35,10 @@
 {%- from 'textarea/macro.njk' import textarea %}
 {%- from 'warning-callout/macro.njk' import warningCallout %}
 
-{% block pageTitle %}{% if errors == true or (errors | length) %}Error: {% endif %}{% if pageName %}{{ pageName }} – {% endif %}{{ serviceName }} – NHS{% endblock %}
+{% block pageTitle -%}
+  {%- if errors == true or (errors | length) %}Error: {% endif -%}
+  {%- if pageName %}{{ pageName }} – {% endif -%}
+  {{ serviceName }} – NHS
+{%- endblock %}
 
 {% set assetPath = '/nhsuk-frontend/assets' %}
--- a/lib/templates/prototype-kit-template.njk
+++ b/lib/templates/prototype-kit-template.njk
@@ -35,6 +35,6 @@
 {%- from 'textarea/macro.njk' import textarea %}
 {%- from 'warning-callout/macro.njk' import warningCallout %}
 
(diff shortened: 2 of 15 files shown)


frankieroberto:
@edwardhorsford @colinrotherham @paulrobertlloyd (and others): thoughts on this? 

Particularly welcome any other ideas for naming the `errorsPresent` variable.

edwardhorsford:
> @edwardhorsford @colinrotherham @paulrobertlloyd (and others): thoughts on this?
> 
> Particularly welcome any other ideas for naming the `errorsPresent` variable.

If I were doing this in my service, I'd probably do it on the basis of the truthyness of an `errors` object. But perhaps for novice users `errorsPresent` is clearer. _Could_ support both I guess.

colinrotherham:
Brilliant, love it! More nice things from that other prototype kit 😎

What about ~`errorsPresent`~ `errorList` as an array?

```njk
{% block pageTitle -%}
  {{ "Error: " if errorList | length }}{{ pageName }} – {{ serviceName }} – NHS
{%- endblock %}
```

Then it matches the error summary param name:

```njk
{% if errorList | length %}
  {{ errorSummary({
    titleText: "There is a problem",
    errorList: errorList
  }) }}
{% endif %}
```

frankieroberto:
@colinrotherham @edwardhorsford naming the errors variable is tricky as there's so many different ways to code error validation within prototypes (plus the option of not bothering at all)!

In the kit itself, the password page sets `error = "wrong-password"` in the routes file.

I’ve also used both `error` and `errors` in my own prototypes 🤦

Another option is to be more explicit and use `addErrorPrefixToTitle` or something?

colinrotherham:
>I’ve also used both `error` and `errors` in my own prototypes 🤦

I'm happy with `error` and `errors` too btw, also used that in another service

```njk
{% if errors | length %}
  {{ errorSummary({
    titleText: "There is a problem",
    errorList: errors
  }) }}
{% endif %}
```

frankieroberto:
@colinrotherham feels like the logic for displaying the error prefix should be based on truthiness rather than `length` though, that way it’s more flexible? eg you might have a hardcoded error page where the error summary doesn't use variables at all?

If you have an empty array you could use something like:

```njk
{% set errorPrefix = (errors | length) %}
```

colinrotherham:
I've used the [`length` filter](https://github.com/mozilla/nunjucks/blob/v3.2.4/nunjucks/src/filters.js#L211-L231) because it only passes on truthy values:

1. Returns `0` (falsy) for `undefined`
1. Returns `0` (falsy) for a map with zero `.size`
1. Returns `0` (falsy) for an array with zero `.length`
1. Returns `0` (falsy) for an object with zero `Object.keys()`

We could check for both `error` or `errors` if you like?

```njk
{% block pageTitle -%}
  {{ "Error: " if error or errors | length }}{{ pageName }} – {{ serviceName }} – NHS
{%- endblock %}
```

Then your `error = "wrong-password"` example works and @edwardhorsford can use `errors` as objects or arrays

frankieroberto:
> I've used the [`length` filter](https://github.com/mozilla/nunjucks/blob/v3.2.4/nunjucks/src/filters.js#L211-L231) because it only passes on truthy values:
> 
> 1. Returns `0` (falsy) for `undefined`
> 2. Returns `0` (falsy) for a map with zero `.size`
> 3. Returns `0` (falsy) for an array with zero `.length`
> 4. Returns `0` (falsy) for an object with zero `Object.keys()`

Nice. I think the only downside with `| length` is that it'll return false if you pass it `true`.

How about we go with a single `errors` variable but do:


```njk
{% block pageTitle %}
  {% if errors or (errors | length) %}Error: {% endif %}{% if pageName %}{{ pageName }} – {% endif %}{{ serviceName }} – NHS
{% endblock %}
```

...that way you could could set `errors` to any of:

* `true`
* a string: `"Enter name"`
* an array: `["Enter name"]`
* an object: `{"name": "Enter name"}`

?

frankieroberto:
@colinrotherham @edwardhorsford have updated it to:

```njk
{% block pageTitle -%}
  {%- if errors == true or (errors | length) %}Error: {% endif -%}
  {%- if pageName %}{{ pageName }} – {% endif -%}
  {{ serviceName }} – NHS
{%- endblock %}
```

Look ok? I'll draft some guidance for it now too.

app/views/index.html
```diff
@@ -10,10 +10,14 @@
...
-  {{ serviceName }}
-{% endblock %}
+<!--
+  Use this to set the page name which will appear at the
+  start of the <title> tag, followed by the service name.
+  This can be set to false on a homepage, where only the
+  service name is needed. You can also reuse this variable
+  within the <h1>, where they should be the same.
+-->
+{% set pageName = false %}
```
colinrotherham: Should probably use `""` as the default for strings?

frankieroberto: This isn't setting a default, it's setting the pageName as false for the homepage as that doesn't need a page name in the title, just the service name.

However it’s a bit redundant as `undefined` would also work so the variable could just not be set at all? I'll remove it.

frankieroberto: Oh wait it's also I guess explaining the feature if you use this page as a template to create new pages from (via copy-and-paste). So maybe it should be an empty string.

frankieroberto:
```suggestion
{% set pageName = "" %}
```

app/views/index.html
```diff
@@ -10,10 +10,14 @@
...
-<!-- Set the page title -->
-{% block pageTitle %}
-  {{ serviceName }}
-{% endblock %}
+<!--
+  Use this to set the page name which will appear at the
+  start of the <title> tag, followed by the service name.
+  This can be set to false on a homepage, where only the
+  service name is needed. You can also reuse this variable
+  within the <h1>, where they should be the same.
```
frankieroberto:
```suggestion
<!--
  Use this to set the page name which will appear at the
  start of the <title> tag, followed by the service name.
  This is often not needed on the homepage, where only the
  service name is needed. You can also reuse this variable
  within the <h1>, where they are the same.
```

app/views/index.html
frankieroberto: Have switched it to an empty string and updated the comment.

frankieroberto:
Draft guidance for this feature here: https://github.com/nhsuk/nhsuk.service-manual.prototype-kit.docs/pull/270

lib/templates/prototype-kit-template.njk
```diff
@@ -35,4 +35,10 @@
 {%- from 'textarea/macro.njk' import textarea %}
 {%- from 'warning-callout/macro.njk' import warningCallout %}
 
+{% block pageTitle -%}
+  {%- if errors == true or (errors | length) %}Error: {% endif -%}
```
colinrotherham: Do we need to worry about `==` (double not triple) being triggered for truthy things?

* https://github.com/mozilla/nunjucks/issues/750

colinrotherham (approved):

lib/templates/prototype-kit-template.njk
frankieroberto: Good point, let's make `=== true`

frankieroberto:
```suggestion
  {%- if errors === true or (errors | length) %}Error: {% endif -%}
```

colinrotherham: Does triple equals definitely work? Didn't test it 😆

frankieroberto: It did, I tested it.

PR #548, opened 2025-06-24
Upgrade to NHS frontend v10

frankieroberto:
This will be a breaking change release of the NHS prototype kit.


app.js
```diff
@@ -162,10 +163,11 @@ exampleTemplatesApp.set('view engine', 'html');
...
 app.set('trust proxy', 1);
 
-// Middleware to serve static assets
+// Use public folder for static assets
 app.use(express.static(path.join(__dirname, 'public')));
-app.use('/nhsuk-frontend', express.static(path.join(__dirname, 'node_modules/nhsuk-frontend/packages')));
-app.use('/nhsuk-frontend', express.static(path.join(__dirname, 'node_modules/nhsuk-frontend/dist')));
+
+// Use assets from NHS frontend
+app.use('/nhsuk-frontend', express.static(path.join(__dirname, 'node_modules/nhsuk-frontend/dist/nhsuk')));
```
frankieroberto, line 170: Really I think we should only be directly serving the `/assets` folder in `nhsuk-frontend` and not the whole parent folder. However currently we rely on this to also copy across the `nhsuk-frontent.min.js` compiled JavasScript file.

We could either switch to just copying that one extra file across (from the Gulpfile), or bundle the JavaScript from source by importing it into `app/assets/javascript/main.js`, which I think would probably be better?

```suggestion
// Use assets from NHS frontend
app.use('/nhsuk-frontend/assets', express.static(path.join(__dirname, 'node_modules/nhsuk-frontend/dist/nhsuk/assets')));
```

frankieroberto:
@anandamaryon1 @colinrotherham I've now updated this to use the latest `internal.3` release - only changes were to the footer params, and adding the notification banner import.

Let me know if you spot anything I've missed!

app.js
frankieroberto: We've decided to keep this as-is for now, serving the whole of `/dist/nhsuk` instead of just the assets subfolder, as it's simpler, means less changes for anyone updating prototypes, and doesn’t cause any particular issues.

Thanks to @colinrotherham and @paulrobertlloyd for the advice!

Possibly something we revisit in future.

colinrotherham (approved):

PR #557, opened 2025-06-26
Use an available port if something is running on the default one

### LINKED ISSUE #114: Automatically use a different port if something is running on port 3000 already
Not a hot burning issue but something that caught me out.

I have a few prototypes for different parts of the programme and across NHS Digital.

Occasionally I find myself running two prototypes that either connect together or I need to evaluate against each other.

Doing this I noticed that you can't run two versions of the prototype kit at once.

It appears initially that you can (one running on port :3000 and :3001 but actually you are only getting port :2000 being proxied.

## Prototype one

image

## Prototype two 
image

The proxy URLs are the different, but the proxied port is not. You see two prototypes running but only one is. This is the same in the browser. 2000, 3000 and 3001 are all the same prototype.

I could change this in one prototype manually to not run on 2000 in `app/config.js`, but I was wondering how easy it would be to run a check on 2000 and if used, run on 2001?

frankieroberto:
This checks whether the port (specified in either `config.js` or as a `PORT` environment variable) is currently available, and if not, asks the user whether to pick a new one. If they answer yes, an available port is selected instead.

This allows multiple prototypes to be more easily run at the same time.

Inspired by from [`findAvailablePort`](https://github.com/alphagov/govuk-prototype-kit/blob/main/lib/utils/index.js#L95-L131) in the GOVUK Prototype Kit but switching out [`inquirer`](https://www.npmjs.com/package/inquirer) for the more recent [`@inquirer/prompts`](https://www.npmjs.com/package/@inquirer/prompts).

I’m only using it from the `gulpfile.js` rather than in `app.js` as it should only be used in local development mode - if you’re running `npm start` on a server like Heroku then it’s better to immediately error if the specified PORT is already in use.

Fixes #114

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -85,10 +85,10 @@ function startNodemon(done) {
   });
 }
 
-function setAvailablePort(done) {
+async function setAvailablePort() {
   const defaultPort = parseInt(process.env.PORT) || config.port;
 
-  findAvailablePort(function(port) {
+  return findAvailablePort(function(port) {
     process.env.PORT = port
     done()
   }, { defaultPort })
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -15,7 +15,7 @@ const config = require('./app/config');
 const { findAvailablePort } = require('./lib/utils');
 
 // Set configuration variables
-const port = parseInt(process.env.PORT || config.port, 10);
+const port = parseInt(process.env.PORT || config.port, 10) || 2000;
 
 // Delete all the files in /public build directory
 function cleanPublic() {
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -9,10 +9,10 @@ const gulpSass = require('gulp-sass')
 const dartSass = require('sass-embedded')
 const nodemon = require('gulp-nodemon');
 const PluginError = require('plugin-error')
-const findAvailablePort = require('./lib/utils/find-available-port');
 
 // Local dependencies
 const config = require('./app/config');
+const findAvailablePort = require('./lib/utils/find-available-port');
 
 // Delete all the files in /public build directory
 function cleanPublic() {
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -14,6 +14,9 @@ const PluginError = require('plugin-error')
 const config = require('./app/config');
 const { findAvailablePort } = require('./lib/utils');
 
+// Set configuration variables
+const port = parseInt(process.env.PORT || config.port, 10);
+
 // Delete all the files in /public build directory
 function cleanPublic() {
   return gulp.src('public', { allowEmpty: true }).pipe(clean());
@@ -58,13 +61,23 @@ function compileAssets() {
 }
 
 // Start nodemon
-function startNodemon(done) {
+async function startNodemon(done) {
+  const availablePort = await findAvailablePort(port);
+  if (!availablePort) {
+    done(new PluginError('startNodemon', `Port ${port} in use`));
+    return;
+  }
+
+  process.env.PORT = availablePort;
+  process.env.WATCH = 'true'
+
   const server = nodemon({
     script: 'app.js',
     stdout: true,
     ext: 'js',
     quiet: false,
   });
+
   let starting = false;
 
   const onReady = () => {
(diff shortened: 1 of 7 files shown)


frankieroberto:
Thinking about it, potentially `portscanner`, `@inquirer/prompts` and `browser-sync` could move to `devDependencies` as they’re not needed when running in 'production' mode (ie on Heroku or Render)?

gulpfile.js
```diff
@@ -9,13 +9,11 @@ const gulpSass = require('gulp-sass')
 const dartSass = require('sass-embedded')
 const nodemon = require('gulp-nodemon');
 const PluginError = require('plugin-error')
+const findAvailablePort = require('./lib/utils/find-available-port');
 
 // Local dependencies
 const config = require('./app/config');
```
colinrotherham: Super minor one! Should this be listed under local dependencies?

```patch
- const findAvailablePort = require('./lib/utils/find-available-port');

  // Local dependencies
  const config = require('./app/config');
+ const findAvailablePort = require('./lib/utils/find-available-port');
```

gulpfile.js
```diff
@@ -87,16 +85,28 @@ function startNodemon(done) {
   });
 }
 
+function setAvailablePort(done) {
+  const defaultPort = parseInt(process.env.PORT) || config.port;
+
+  findAvailablePort(function(port) {
+    process.env.PORT = port
+    done()
+  }, {defaultPort: defaultPort})
```
colinrotherham: Probably something ESLint auto fix one day

```suggestion
  }, { defaultPort })
```

gulpfile.js
```diff
@@ -87,16 +85,28 @@ function startNodemon(done) {
   });
 }
 
+function setAvailablePort(done) {
+  const defaultPort = parseInt(process.env.PORT) || config.port;
+
+  findAvailablePort(function(port) {
```
colinrotherham: Shall we return a promise here so Gulp can handle errors for us?

```suggestion
async function setAvailablePort() {
  const defaultPort = parseInt(process.env.PORT) || config.port;

  return findAvailablePort(function(port) {
```

lib/utils/find-available-port.js
```diff
@@ -0,0 +1,47 @@
...
+
+      if (defaultPort === availablePort) {
+        // Default port is free, return it via the callback
+        callback(defaultPort);
+      } else {
+        console.error(`ERROR: Port ${defaultPort} in use - you may have another prototype running.\n`); // eslint-disable-line no-console
+
+        inquirer.confirm({
+          message: 'Change to an available port?',
+        }).then((answeredYes) => {
```
colinrotherham: Mind if we flatten this a bit to avoid the callbacks and promise chains?

Somewhat simplified, but can probably reduce it to:

```js
const availablePort = await portScanner.findAPortNotInUse()

if (defaultPort === availablePort || await inquirer.confirm()) {
  return availablePort
}

// throw
```

colinrotherham (commented):
Love seeing this functionality, reminds me we need it for NHS.UK frontend!

gulpfile.js
frankieroberto: Good point!

lib/utils/find-available-port.js
frankieroberto: I got a bit lost with all the chains of callbacks, promises and async functions too...

Would you mind taking a look and committing any suggestions to make it more readable and conventional? (Maybe we should use Promises throughout?)

colinrotherham (approved):

edwardhorsford (changes requested):
I've tested this with an existing prototype running, and when I answer 'yes' that I want to change ports, it crashes. the Error is `Error: No open port found`.

Chatting with @colinrotherham we think this might be because the kit actually runs on two ports by default - 2000 and 3000 - and this is perhaps only checking if one of those is free?

edwardhorsford (approved):
With @colinrotherham's fix this seems to be working now. Great stuff!

colinrotherham:
Thanks @edwardhorsford, glad you found it

Yeah the start/end syntax was incorrect, but working now:

```patch
-   return findAPortNotInUse([startPort, startPort + 1000], host);
+   return findAPortNotInUse(startPort, startPort + 1000, host);
```

I've added default port should someone remove the config line too

PR #602, opened 2025-08-28
Exclude app code from linters by default

### LINKED ISSUE #601: Version 7 runs linting on app code after update
# Bug Report

## What is the issue?

When upgrading to version 7 the new `app/.eslintrc.js` config is excluded by the instructions

>…delete everything apart from the **app** and **.git** folder.

Assuming we want to ignore Prettier too?

As above, without `app/.eslintrc.js` ESLint will start linting files inside **app**

We should move the ESLint ignore rules to the root config so it's not missed

colinrotherham:
## Description

Fixes #601 and adds a default editor config file

## Checklist

- [x] CHANGELOG entry


frankieroberto (approved):
Thanks. Have also added a line to the changelog about adding the `.editorconfig` file.

PR #613, opened 2025-09-05
Use Node.js built in `.env` support

### LINKED ISSUE #600: Consider removing dotenv
Should we remove `dotenv` from the kit as a dependency? Or make it a devDependency?

Currently it's included with the kit, including in production mode, but most hosts like Heroku or Railway have their own way to configure environment variables.

Whilst `.env` files are useful, we also have other ways to configure things:

`app/config.js` - eg sevice name and port can be set here
`app/locals.js` - for anything you want to be able to access in Nunjucks views

If `dotenv` were removed, people could still use it by either installing the command line tool ([dotenv-cli](https://www.npmjs.com/package/dotenv-cli) globally and using `dotenv npm run watch`, or just adding it back in to their `package.json` manually.

Alternatively, we could keep it in but move it to `devDependencies`?

Either way, we can continue to keep `.env` in the `.gitignore` to prevent the files from being accidentally committed to GitHub.

colinrotherham:
## Description

This PR replaces `dotenv` with Node.js [`process.loadEnvFile()`](https://nodejs.org/api/process.html#processloadenvfilepath)

I've added a safety check as this feature requires Node.js v20.12.0 or later

Fixes #600

## Checklist

- [x] CHANGELOG entry


app.js
```diff
@@ -6,21 +6,26 @@ const {
...
-})
+/**
+ * Add environment variables from .env file
+ * (Requires Node.js v20.12.0 or later)
+ *
+ * @see {@link https://nodejs.org/api/process.html#processloadenvfilepath}
+ */
+if ('loadEnvFile' in process) {
+  process.loadEnvFile()
+}
```
frankieroberto: Should we only do this from the `watch` command, so that it never happens on a remote host like Heroku/Railway/etc?

colinrotherham: Some teams have copied **.env** files into Docker containers for deployment

I'm already wary that this PR drops `.env` support for Node.js v20.9.0–20.11.1

```json
  "engines": {
    "node": "^20.9.0 || ^22.11.0"
  }
```

That might be a reason to put this change in v8

frankieroberto: ok I'll hold this for the next release.

frankieroberto: I've moved this to the `gulpfile.js` now, to simplify the `app.js` and make sure it’s not used in production mode.

colinrotherham: Ah didn't we say some teams might use it in production?

colinrotherham:
@frankieroberto What if `.env` exists but it's not formatted correctly?

We don't want to suppress the error

colinrotherham:
Great minds, you've sorted it 🙌

frankieroberto:
I’ve copied this into the other monster PR: https://github.com/nhsuk/nhsuk-prototype-kit/pull/688/commits/2da579b20c28735af884f8b1cc1d4386c137a542

frankieroberto:
Closing this, as we’ve made the change upstream on https://github.com/nhsuk/nhsuk-prototype-kit-package/ instead.

PR #628, opened 2025-09-10
Open preview in a new browser tab instead of an preview

frankieroberto:
This defaults the prototype to running in a new browser tab, instead of a tab within Codespaces itself.

The advantage to this is that if you refresh it using Control-R, it'll only refresh the browser tab and not the whole of codespaces. It also makes it a bit easier to test the responsiveness by resizing the window.

The downside is that depending on your browser settings you might get a popup warning first:

Screenshot 2025-09-10 at 10 30 03


edwardhorsford (approved):
Awesome! Tested in codespaces and it automatically opened the preview in a new tab.

PR #629, opened 2025-09-10
Fix Browsersync in GitHub Codespaces

### LINKED ISSUE #622: BrowserSync doesn't work in Codespaces
We’ve so far not been able to get BrowserSync working in Codespaces (@paulmsmith and I had a go but couldn’t figure it out).

colinrotherham:
## Description

This PR makes sure Browsersync works in GitHub Codespaces to close https://github.com/nhsuk/nhsuk-prototype-kit/issues/622

Since Browsersync prepends its own **browser-sync-client.js** script using [**templates/connector.tmpl**](https://github.com/BrowserSync/browser-sync/blob/master/packages/browser-sync/templates/connector.tmpl) we can follow the URL building code in [**lib/connect-utils.js**](https://github.com/BrowserSync/browser-sync/blob/dc74bc0a6bcbe70eea5f2f94b524b2902436efb1/packages/browser-sync/lib/connect-utils.js#L176-L186) to make sure we avoid custom port numbers in GitHub Codespaces.

For example:

### ❌ Browsersync as a proxy
When configured via [`proxy`](https://browsersync.io/docs/options#option-proxy) the Browsersync socket request includes the port:

```js
___browserSync___.socketUrl = 'http://' + location.hostname + ':3000/browser-sync';
```

### ✅ Browsersync as a server
When configured via [`server`](https://browsersync.io/docs/options#option-server) the Browsersync socket request uses the same host as the application:
```js
___browserSync___.socketUrl = '' + location.host + '/browser-sync';
```

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,3 +1,5 @@
+const { join } = require('node:path')
+
 // External dependencies
 const browserSync = require('browser-sync')
 const gulp = require('gulp')
@@ -5,6 +7,7 @@ const babel = require('gulp-babel')
 const clean = require('gulp-clean')
 const nodemon = require('gulp-nodemon')
 const gulpSass = require('gulp-sass')
+const { createProxyMiddleware } = require('http-proxy-middleware')
 const PluginError = require('plugin-error')
 const dartSass = require('sass-embedded')
 
@@ -127,14 +130,24 @@ async function startBrowserSync(done) {
 
   browserSync.init(
     {
-      proxy: `localhost:${proxyPort}`,
       port: proxyPort + 1000,
       ui: false,
       files: ['app/views/**/*.*', 'lib/example-templates/**/*.*'],
       ghostMode: false,
       open: false,
       notify: true,
-      watch: true
+      watch: true,
+
+      // Proxy to Node.js server
+      middleware: createProxyMiddleware({
+        changeOrigin: true,
+        target: `http://localhost:${proxyPort}`
+      }),
+
+      // Serve static assets
+      server: {
+        baseDir: join(__dirname, 'public')
+      }
     },
     done
   )
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "gulp-nodemon": "^2.5.0",
     "gulp-rename": "^2.1.0",
     "gulp-sass": "^6.0.1",
+    "http-proxy-middleware": "^3.0.5",
     "lodash": "^4.17.21",
     "nhsuk-frontend": "^10.0.0",
     "nunjucks": "^3.2.4",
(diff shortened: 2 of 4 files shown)


gulpfile.js
```diff
@@ -127,14 +130,24 @@ async function startBrowserSync(done) {
 
   browserSync.init(
     {
-      proxy: `localhost:${proxyPort}`,
```
colinrotherham, line 130: We did actually need this, but sadly it adds `:3000` to the socket URL request 😭

gulpfile.js
```diff
@@ -127,14 +130,24 @@ async function startBrowserSync(done) {
...
       open: false,
       notify: true,
-      watch: true
+      watch: true,
+
+      // Proxy to Node.js server
+      middleware: createProxyMiddleware({
+        changeOrigin: true,
+        target: `http://localhost:${proxyPort}`
+      }),
```
colinrotherham, line 145: We can be cheeky and add middeware that intercepts all requests and _proxies_ them

i.e. We get a proxy without the [`proxy` option](https://browsersync.io/docs/options#option-proxy) 😎

Thus avoiding the port `:3000` in the socket URL request

gulpfile.js
```diff
@@ -127,14 +130,24 @@ async function startBrowserSync(done) {
...
+
+      // Serve static assets
+      server: {
+        baseDir: join(__dirname, 'public')
+      }
```
colinrotherham, line 150: This is kinda redundant but Browsersync needs the [`server` option](https://browsersync.io/docs/options#option-server) to [dodge the port number logic](https://github.com/BrowserSync/browser-sync/blob/dc74bc0a6bcbe70eea5f2f94b524b2902436efb1/packages/browser-sync/lib/connect-utils.js#L176-L186)

What's the problem? If you kill the Express.js application the assets are still served 🤷‍♂️

package.json
```diff
@@ -38,6 +38,7 @@
     "gulp-nodemon": "^2.5.0",
     "gulp-rename": "^2.1.0",
     "gulp-sass": "^6.0.1",
+    "http-proxy-middleware": "^3.0.5",
```
frankieroberto, line 41: Feels like this should be in `devDependencies`, so that it can be skipped on Heroku and other hosts running prototypes in `NODE_ENV=production` mode?

colinrotherham: For now though, it's alongside `browsersync` and `gulp-nodemon`?

frankieroberto: Ah yeah. 🤦  Spun out as a separate issue: #632

edwardhorsford:
Had a play with this locally (not Codespaces) and it seems to all work - opens fine and auto reloads.

frankieroberto (approved):
Nice, thanks!

PR #631, opened 2025-09-10
Turn off minimap by default in Codespaces

frankieroberto:
I think the [Minimap](https://code.visualstudio.com/docs/getstarted/userinterface#_minimap) feature is probably a bit distracting for new users?

(It can still be turned on again in the interface)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -33,6 +33,8 @@
     "vscode": {
       // editor settings
       "settings": {
+        // Turn off minimap by default, as it's distracting for new users
+        "editor.minimap.enabled": false,
         // uncomment the following lines to hide files not needed to update content
         // "files.exclude": {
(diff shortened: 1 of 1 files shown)


colinrotherham:
[Zen mode](https://code.visualstudio.com/docs/getstarted/userinterface#_zen-mode)?

frankieroberto:
@colinrotherham I think Zen mode is maybe going a bit too far, but let’s see. Would you agree that Minimap is distracting, at least for beginners?

@edwardhorsford happy with turning this off by default too?

edwardhorsford:
As a person who uses the mini-map, I'm unsure about turning this off.

frankieroberto:
@edwardhorsford you would still be able to turn it on again, this is just the initial default setting. 

But happy to leave this for now, and see if it throws anyone else on the courses.

PR #645, opened 2025-09-19
Fix nested query strings not expanding

### LINKED ISSUE #644: Nested query params no longer save to session correctly
It's no longer possible to use a nested query params and have that save correctly to an expanded object.

If you visit `localhost:3000/?zip[baz]=buz`, what should save is:

Correct:
`{ zip = { baz: "buz" }}`

What gets saved currently:
`{ "zip[baz]" = "buz" }`

This used to work in the kit but no longer works. Chatting with @colinrotherham we think this is related to the upgrade to Express v5.

>req.query
The req.query property is no longer a writable property and is instead a getter. The default query parser has been changed from “extended” to “simple”.

edwardhorsford:
## Description
Fixes #644

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 :wrench: **Fixes**
 
 - Fix Browsersync in Codespaces
+- Fix nested query strings not saving correctly (PR, Fixes [#644](https://github.com/nhsuk/nhsuk-prototype-kit/issues/644))
 
 ## 7.0.1 - 5 September 2025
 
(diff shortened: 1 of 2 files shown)


colinrotherham (approved):
Looks good

Does it definitely work when `app.set()` comes after the `app.use()` calls?

Just in case other dependencies are checking for `extended` parsing

@frankieroberto Noticed we stopped linking to PRs and issues in the changelog, do you want to continue with that or doesn't matter?

frankieroberto (approved):

PR #663, opened 2025-10-02
Add filter for formatting NHS numbers

### LINKED ISSUE #654: Add an NHS number formatting filter
We should add a Nunjucks filter which formats NHS numbers according to the NHS style guide:

> Write the NHS number as 3 groups of numbers, with a single space between them, like this: [485 777 3456](tel:485 777 3456). (This is an example or test number.)
> 
> The format is 3 3 4: 2 groups of 3 digits followed by 4 digits.

frankieroberto:
This adds a built-in filter for formatting NHS numbers using spaces.

The number format is:

> The format is 3 3 4: 2 groups of 3 digits followed by 4 digits.

This number format makes it easier for people and assistive technologies to read. It also makes it less likely that people will make mistakes.

See https://service-manual.nhs.uk/content/a-to-z-of-nhs-health-writing#nhs-number
 
Resolves #654

### DIFF
--- a/lib/filters/format-nhs-number.js
+++ b/lib/filters/format-nhs-number.js
@@ -3,7 +3,8 @@
  *
  * The format is 3 3 4: 2 groups of 3 digits followed by 4 digits.
  *
- * For example, "4857773456" becomes "485 777 3456". (This is an example or test number.)
+ * For example, "4857773456" becomes "485 777 3456".
+ * (This is an example or test number.)
  *
  * This number format makes it easier for people and assistive technologies
  * to read. It also makes it less likely that people will make mistakes.
--- a/lib/filters/format-nhs-number.js
+++ b/lib/filters/format-nhs-number.js
@@ -1,6 +1,8 @@
 /**
  * Display an NHS number formatted with spaces
  *
+ * The format is 3 3 4: 2 groups of 3 digits followed by 4 digits.
+ *
  * This number format makes it easier for people and assistive technologies
  * to read. It also makes it less likely that people will make mistakes.
  *
--- a/lib/filters/format-nhs-number.js
+++ b/lib/filters/format-nhs-number.js
@@ -3,6 +3,8 @@
  *
  * The format is 3 3 4: 2 groups of 3 digits followed by 4 digits.
  *
+ * For example, "485 777 3456". (This is an example or test number.)
+ *
  * This number format makes it easier for people and assistive technologies
  * to read. It also makes it less likely that people will make mistakes.
  *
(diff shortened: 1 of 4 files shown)


edwardhorsford:
Here's mine:
```js
/**
 * Format NHS number with spaces (3-3-4 format)
 *
 * @param {string|number} input - NHS number to format
 * @returns {string} Formatted NHS number or original input if invalid
 */
const formatNhsNumber = (input) => {
  if (!input) return ''
  const nhsStr = input.toString().replace(/\s/g, '')

  if (!/^\d{10}$/.test(nhsStr)) {
    return input
  }

  return `${nhsStr.slice(0, 3)} ${nhsStr.slice(3, 6)} ${nhsStr.slice(6)}`
}
```

lib/filters/format-nhs-number.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * Display an NHS number formatted with spaces
+ *
+ * This number format makes it easier for people and assistive technologies
+ * to read. It also makes it less likely that people will make mistakes.
+ *
+ * See https://service-manual.nhs.uk/content/a-to-z-of-nhs-health-writing#nhs-number
```
edwardhorsford, line 12: This description could include an example of the output format?

frankieroberto: Done in 829bc6f9a782f7d97bc5c6a433f4bc2e3163fe31

frankieroberto: Although maybe an example is clearer

frankieroberto: Added the example in 48fd848faf26cfcb18dd7c6c7191eaea3aa99a28

edwardhorsford: Yeah I sort of meant a before and after. I'm hopeful we'll soon change our guidance to using example numbers that start with `999` which is more unambiguously a test number.

frankieroberto: Added the 'before' now too. That seem ok?

Can switch to 999 number if prefer it. I'd thought that "485 777 3456" was also a test number?

sarawilcox:
We're proposing to change the old test number, Frankie, as it seems to be causing some confusion. Latest proposed number is 999 123 4567.

frankieroberto:
> We're proposing to change the old test number, Frankie, as it seems to be causing some confusion. Latest proposed number is 999 123 4567.

Thanks - have switched to that number in the examples.

edwardhorsford (approved):
Have tested locally with null, undefined, number, strings, all seems to work great and as expected.

PR #688, opened 2025-10-21
Use NHS Prototype kit package

### LINKED ISSUE #63: Extract core of the prototype kit into an npm package
In order to [make the kit easier to update](https://github.com/nhsuk/nhsuk-prototype-kit-package/issues/60), we can extra most of the kit (the bits that users shouldn't touch) into an node package, which can be versioned and released via npm, and then updated via Dependabot or by updating `package.json`.

See also:

* #60

frankieroberto:
This adds a dependency on a new `nhsuk-prototype-kit` package, where the code in `lib/` has been moved to.

This then simplifies the `app.js` file, and will make updates to the code a lot easier.

As part of this change, this repo would then become versionless, in recognition of its role as a 'template repositroy', where the 'Use this template' button always creates a copy from the latest version in `main`.

The `nhsuk-prototype-kit` package version numbering would start at `8.0.0`, picking up the numbering from this repo (currently on `7.x`).

Resolves nhsuk/nhsuk-prototype-kit-package#63


frankieroberto:
@colinrotherham I think this is mergeable, now `8.0.0` is released! 🎉

PR #694, opened 2025-11-10
Add Dependabot groups with monthly checks

colinrotherham:
## Description

This PR adds Dependabot groups to reduce package update noise

I've removed the Sass update restriction so we can use `@dependabot` comments to ignore updates instead. We might accidentally miss important (or incompatible) changes without it

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
     "lodash": "^4.17.21",
     "nhsuk-frontend": "^10.1.0",
     "nunjucks": "^3.2.4",
-    "path": "^0.12.7",
     "plugin-error": "^2.0.1",
     "portscanner": "^2.2.0",
     "sass-embedded": "1.77.5"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -42,7 +42,6 @@
     "lodash": "^4.17.21",
     "nhsuk-frontend": "^10.1.0",
     "nunjucks": "^3.2.4",
-    "path": "^0.12.7",
```
colinrotherham, line 45: Not sure why we used `path` but we updated to `node:path` recently to avoid it

I hadn't actually realised we used an alternative from the registry

frankieroberto:
Closing this, as it’ll conflict with #688 - after that’s merged, most of the dependabot PRs will get raised at http://github.com/nhsuk/nhsuk-prototype-kit-package/ instead.

colinrotherham:
@frankieroberto The groups are a bit help though, less noise

PR #695, opened 2025-11-10
Upgrade to Node.js 24 with backwards compatibility

colinrotherham:
## Description

This PR adds Node.js 24 to the **package.json** supported `engines` range

```patch
  "engines": {
 -  "node": "^20.9.0 || ^22.11.0"
+   "node": "^20.9.0 || ^22.11.0 || ^24.11.0"
  }
```

I've also updated PR checks and config files for Browserslist and nvm

## Checklist

- [x] CHANGELOG entry


frankieroberto (approved):
Thanks!

Let's copy this across to https://github.com/nhsuk/nhsuk-prototype-kit-package too.

colinrotherham:
Just flagging that the current **.nvmrc** file in this repo is set to Node.js 20 😬

We want it in both places yeah?

colinrotherham:
Ah yes you approved this, thanks

frankieroberto:
Oh good point. I wonder if we should remove `.nvmrc` from version control by adding it to `.gitignore`?

PR #530, opened 2025-05-23
Switch to StandardJS for javascript linting

frankieroberto:
Fixes #533

## Checklist

- [ ] Tested against our [testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
- [ ] CHANGELOG entry

PR #696, opened 2025-11-10
Upgrade to ESLint v9

colinrotherham:
## Description

This PR upgrades to ESLint v9 now that all packages are compatible

Previously in https://github.com/nhsuk/nhsuk-prototype-kit/pull/530 we were only able to use ESLint v8

## Checklist

- [ ] CHANGELOG entry


frankieroberto:
@colinrotherham would you mind shifting this work across to target https://github.com/nhsuk/nhsuk-prototype-kit-package instead? That repo contains the kit extracted as an NPM package. It currently has prettier set up, but not ESLint yet.

My current thinking is that once the npm package is released, this template repo would be updated to use it, and would no longer itself containing any linting or tests. (Although users could choose to add them if they prefer). See #688 for work in progress.

frankieroberto:
@colinrotherham I’ve had a go at doing this at https://github.com/nhsuk/nhsuk-prototype-kit-package/pull/41

Mind if I close this for now? Assuming we can upgrade to v8 fairly soon, this PR would become redundant as the template repo wouldn’t have linting set up.

colinrotherham:
Up to you. There's still value in merging it

PR #700, opened 2025-11-10
Fix ES module imports in browser JavaScript

### LINKED ISSUE #699: Error importing from NHS.UK frontend
# Bug Report

## What is the issue?

ES module imports in browser JavaScript files are transformed to Node.js `require()` calls

This throws the following error in a browser:

```console
ReferenceError: Can't find variable: require
```

Since we don't use a bundler (only `gulp-babel`), we should configure Babel to preserve modules

The HTML JavaScript snippet that imports NHS.UK frontend is unaffected

Add the following code to `app/assets/javascript/main.js`

```mjs
import { createAll } from '/nhsuk-frontend/nhsuk-frontend.min.js'
```

Notice the contents of `public/js/main.js` becomes:

```mjs
"use strict";

var _nhsukFrontendMin = require("/nhsuk-frontend/nhsuk-frontend.min.js");//# sourceMappingURL=main.js.map
```

N/A

colinrotherham:
## Description

This PR closes https://github.com/nhsuk/nhsuk-prototype-kit/issues/699 by configuring Babel to preserve modules

Without this fix, Babel will transform `import` to `require()` ready for bundling

We don't have a bundler

## Checklist

- [x] CHANGELOG entry


frankieroberto:
Closing this, as it’ll be made redundant by #688 which no longer uses babel.

PR #701, opened 2025-11-11
[SPIKE] Add NHS.UK frontend code autocomplete

colinrotherham:
## Description

This PR adds the TypeScript compiler to automatically discover code autocomplete suggestions etc

Seems a shame not to use the published type declarations from NHS.UK frontend package

E.g. Moving the `initAll()` import from HTML to the **main.js** file

Code autocomplete

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/app.js
+++ b/app.js
@@ -245,7 +245,11 @@ app.use((req, res, next) => {
 })
 
 // Display error
-app.use((err, req, res) => {
+app.use((err, req, res, next) => {
+  if (res.headersSent) {
+    return next(err)
+  }
+
   console.error(err.message)
   res.status(err.status || 500)
   res.send(err.message)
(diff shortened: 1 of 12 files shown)


app.js
```diff
@@ -245,7 +245,11 @@ app.use((req, res, next) => {
 })
 
 // Display error
-app.use((err, req, res) => {
+app.use((err, req, res, next) => {
+  if (res.headersSent) {
+    return next(err)
+  }
+
```
colinrotherham, line 252: I've added this as a little example of why using types can be good

Error middleware checks in Express.js use `Function.length` to check for 4x calling arguments

i.e. We've missed the `next` param here

frankieroberto:
Closing this, as it’d be more to help the development of the kit, rather than for users of the kit itself, as I understand it. Could re-raise on http://github.com/nhsuk/nhsuk-prototype-kit-package/ ?

PR #730, opened 2026-02-06
Remove sass-embedded from dependabot config

frankieroberto:
This no longer needs to be specified as it's not a direct dependency any more, but is a sub-dependency of `nhsuk-prototype-kit`


colinrotherham:
@copilot Can you check if any other dependencies need adding or removing from the Dependabot config?

colinrotherham (approved):

PR #736, opened 2026-02-09
Use default 3000 port for Codespaces

frankieroberto:


frankieroberto:
Tested on Codespaces. Seems to work! 🙌

colinrotherham (approved):

.devcontainer/devcontainer.json
colinrotherham, line 23: This is the command to (hopefully) prevent lockfile changes

```suggestion
  "postCreateCommand": "npm install --no-save",
```

colinrotherham: No idea why the diff isn't working

PR #752, opened 2026-03-23
Add kit and frontend version numbers in footer

### LINKED ISSUE #239: Show version numbers
With the upgrade to v8, we no longer show the prototype kit version number on the index page. This makes it a bit harder to know which version you are on. 

Now that updates are via packages, this might be more confusing too. 

So, I wonder whether we could add the current kit and frontend version somewhere visible, like the home page or in the footer. 

Footer would have the benefit of showing on each page, so might make it easier when debugging / helping people from screenshots.

frankieroberto:
This adds the version number of NHS prototype kit and NHS frontend into the default footer.

Screenshot 2026-03-23 at 13 14 11

Resolves https://github.com/nhsuk/nhsuk-prototype-kit-package/issues/239

This is just the default footer in this template repository, so users can change it if needed.

### DIFF
--- a/app/views/layout.html
+++ b/app/views/layout.html
@@ -26,6 +26,10 @@
 <!-- Edit the footer -->
 <!-- Footer code examples can be found at https://service-manual.nhs.uk/design-system/components/footer -->
 {% block footer %}
+  {% set metaAboutHtml %}
+    <p class="nhsuk-body-s">Built on <a href="https://prototype-kit.service-manual.nhs.uk/updates/"  class="nhsuk-footer__list-item-link">NHS prototype kit {{ prototypeKitVersion }}</a> and <a href="https://github.com/nhsuk/nhsuk-frontend/releases/"  class="nhsuk-footer__list-item-link">NHS.UK frontend {{ nhsukFrontendVersion }}</a></p>
+  {% endset %}
+
   {{ footer({
     meta: {
       items: [
@@ -38,7 +42,7 @@
           text: "Reset data"
         }
       ],
-      html: '<p class="nhsuk-body-s">Built on <a href="https://prototype-kit.service-manual.nhs.uk/updates/">NHS prototype kit ' + prototypeKitVersion + '</a> and <a href="">NHS.UK frontend ' + nhsukFrontendVersion + '</a></p>'
+      html: metaAboutHtml
     }
   }) }}
 {% endblock %}
--- a/app/views/layout.html
+++ b/app/views/layout.html
@@ -37,7 +37,8 @@
           href: "/prototype-admin/reset?returnPage=" + (currentPage | urlencode),
           text: "Reset data"
         }
-      ]
+      ],
+      html: '<p class="nhsuk-body-s">Built on <a href="https://prototype-kit.service-manual.nhs.uk/updates/">NHS prototype kit ' + prototypeKitVersion + '</a> and <a href="">NHS.UK frontend ' + nhsukFrontendVersion + '</a></p>'
     }
   }) }}
 {% endblock %}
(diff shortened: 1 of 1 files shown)


app/views/layout.html
```diff
@@ -37,7 +37,8 @@
           href: "/prototype-admin/reset?returnPage=" + (currentPage | urlencode),
           text: "Reset data"
         }
-      ]
+      ],
+      html: '<p class="nhsuk-body-s">Built on <a href="https://prototype-kit.service-manual.nhs.uk/updates/">NHS prototype kit ' + prototypeKitVersion + '</a> and <a href="">NHS.UK frontend ' + nhsukFrontendVersion + '</a></p>'
```
colinrotherham: Sorry @frankieroberto, these need `nhsuk-footer__list-item-link` to pick up footer link styles

Are you alright [lifting the code from the package](https://github.com/nhsuk/nhsuk-prototype-kit-package/pull/230/changes#diff-f757e211b2c187bc5168804ae445e8bc1a93cb33f4fad2ce9bcd501009db4d59) that we did with @anandamaryon1?

Thanks

frankieroberto: Added the classes, thanks!

I've made the NHS frontend link go to https://github.com/nhsuk/nhsuk-frontend/releases/ rather than for that specific release, as that might encourage users to more easily see whether there are any new versions, and how up-to-date they are? At some point though we'll swap this out for a link to the service manual website, when that has a page on it that lists releases. (@anandamaryon1 says this is in the words?)

colinrotherham: @anandamaryon1 and I were happy with the footer versions we added to the package though?

It keeps the old behaviour too, linking to the version in the link text

frankieroberto: @colinrotherham when we discuss this in the call this morning we thought it might be better to avoid linking to GitHub if possible, and instead link to pages on the prototype kit and service manual websites, as navigating GitHub can be quite confusing for less technical folk.

Happy to hear alternate arguments though!

colinrotherham: No no, that's fine. Thanks for explaining

I just saw this PR for what it was, and wondered what was wrong with the test app version

app/views/layout.html
```diff
@@ -26,6 +26,10 @@
 <!-- Edit the footer -->
 <!-- Footer code examples can be found at https://service-manual.nhs.uk/design-system/components/footer -->
 {% block footer %}
+  {% set metaAboutHtml %}
+    <p class="nhsuk-body-s">Built on <a href="https://prototype-kit.service-manual.nhs.uk/updates/"  class="nhsuk-footer__list-item-link">NHS prototype kit {{ prototypeKitVersion }}</a> and <a href="https://github.com/nhsuk/nhsuk-frontend/releases/"  class="nhsuk-footer__list-item-link">NHS.UK frontend {{ nhsukFrontendVersion }}</a></p>
```
colinrotherham: If you can stick the [`class` attribute before `href`](https://github.com/nhsuk/nhsuk-frontend/blob/main/docs/contributing/coding-standards.md#attribute-order) and remove the double spaces?

```suggestion
    <p class="nhsuk-body-s">Built on <a class="nhsuk-footer__list-item-link" href="https://prototype-kit.service-manual.nhs.uk/updates/">NHS prototype kit {{ prototypeKitVersion }}</a> and <a class="nhsuk-footer__list-item-link" href="https://github.com/nhsuk/nhsuk-frontend/releases/">NHS.UK frontend {{ nhsukFrontendVersion }}</a></p>
```

colinrotherham:
Thanks for this. Shall we turn review apps on for easy reviews?

frankieroberto:
> Thanks for this. Shall we turn review apps on for easy reviews?

I'm getting an error when I try. Might need to be an admin maybe?

colinrotherham (approved):

